PR #2, opened 2022-08-19
feat: Added changelog page

yahu1031:
Added changelogs page

Screenshot 2022-08-19 at 5 03 36 PM


vlidholt:
Will not be included unless we can automate it.

PR #8, opened 2023-02-12
feat: add cURL instructions

### LINKED ISSUE #7: Add documentation details on how to call endpoints using cURL
I noticed that in the documentation are missing the details on how to call the endpoints using cURL (or other API platorms, such as Postman).

Would be nice to add this detail.

davideroffo:
Closes #7 

I've added details on the documentation on how to call serverpod's endpoints using cURL.


vlidholt:
To include this, we should probably include the spec for the whole API, included serializable objects, etc. Hope to do it in the future!

PR #12, opened 2023-04-26
Fix the path of the docker directory

wilsonsilva:
Fixes the path to the Docker directory [docker-local](https://github.com/serverpod/serverpod/tree/9c251e16af05a11a13e97a0f98e3bf302b50358a/tests/serverpod_test_server/docker-local).


vlidholt:
Well spotted. Would have merged it if it was corrected in the latest versions. Just added the fix in the right markdown file now.

PR #17, opened 2023-07-19
Update index.md

mjvanmiddelaar:
Docker compose has been added as a subcommand to the docker executable. 

https://docs.docker.com/engine/reference/commandline/compose/


vlidholt:
Thank you, this was already included in the next version of the docs (which is now released). Very much appreciate the contribution though!

PR #18, opened 2023-07-20
Suggested test guidelines

SandPod:
Suggestion for test guidelines.

There doesn't seem to be any general test guidelines for dart provided by google. So I created this based on common best practices and applied them to the dart test framework.

Does this seem reasonable?
Anything missing?

### DIFF
--- a/docs/09-contribute.md
+++ b/docs/09-contribute.md
@@ -15,6 +15,110 @@ The main [Serverpod repository](https://github.com/serverpod/serverpod) contains
 ### Writing code
 We are very conscious about keeping the Serverpod code base clean. When you write your code, make sure to use `dart format` and that you don't get any errors or lints from `dart analyze`.
 
+### Writing tests
+Automated tests help us keep serverpod stable and simplifies safe contributions. To ensure that your code works as expected we require you to add automated test along with any contribution. By adhering to these guidelines, we can create an effective and helpful test suite.
+
+#### Minimize scope
+A failed test should help a developer quickly pinpoint what has gone wrong, therefore we aim to minimize the scope of each test. An ideal test should only have one reason to fail.
+
+```dart title="Bad"
+  test('Ice cream truck when buying strawberry ice cream returns strawberry ice cream and was discounted', () {
+    final IceCreamTruck truck = IceCreamTruck();
+
+    final iceCream = truck.buyStrawberryIceCream();
+
+    expect(iceCream.type, equals('strawberry'));
+    expect(iceCream.discount, isTrue);
+  });
+```
+
+```dart title="Good"
+  test('Ice cream truck when buying strawberry ice cream should return strawberry ice cream', () {
+    final IceCreamTruck truck = IceCreamTruck();
+
+    final iceCream = truck.buyStrawberryIceCream();
+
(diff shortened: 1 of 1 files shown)


Isakdl:
Over all pretty solid :)

Some thoughts to add to the discussion:

Do we favor comments over explicitly putting the `given` / `when` / `then` in the test description? In the example you have here we are doing both.

This example: `Cat when belly is scratched then scratches back` we have both `when` and `then` but a "silent" `given`. How explicit should we be here? Any reason for not writing `given`?

What about the word `should`? I have had this discussion before and not 100% what I think but there is an argument for that a test expects something. I.E. then something happens, not that it should happen. Very philosophical and subtle difference 😄. But given that we advocate for the Given, When, Then, syntax it might make it more clear if we simply replaced `should` with `then`.

Example: `when added multiplied should return product` vs `when added multiplied then return product`

What do you think?

Isakdl:
Some other things we could think about adding:

Encourage the usage of test builders to construct the objects we want to test. I have started using these for the tests I'm making for the code generator currently.

Avoid mocking as much as possible and push side effects as far up the stack as possible. As this makes the tests much less brittle and independent from the implementation. 

Test file structure, where do we put the test files and what names do we give them? For example in the case of the analyzer I have many test files for a single method as the tests are several thousands lines of code. Do we want some guideline for how to organize our files? 

Test scope, how much code are we testing? Personally I think unit tests that cover large amount of (pure) code is to be preferred, meaning potentially multiple classes are tested indirectly. Perhaps what @SandPod would call module tests. Unit tests that only test a single method deep inside the logic are in my experience pretty useless and gets in the way. There are of course exceptions such as smaller methods that are used in many different places or are critical, I wouldn't mind a test that explicitly verifies a regex for example. 

Abstraction level of tests. I think encouraging a direct and explicit test code is in most cases a benefit. I.E. not hiding expects deep in some hidden method or obscure test setups. As a developer you should be able to more or less understand the test by only looking at a specific test in isolation.

Thoughts? :)

SandPod:
> Over all pretty solid :)
> 
> Some thoughts to add to the discussion:
> 
> Do we favor comments over explicitly putting the `given` / `when` / `then` in the test description? In the example you have here we are doing both.
> 
> This example: `Cat when belly is scratched then scratches back` we have both `when` and `then` but a "silent" `given`. How explicit should we be here? Any reason for not writing `given`?
> 
> What about the word `should`? I have had this discussion before and not 100% what I think but there is an argument for that a test expects something. I.E. then something happens, not that it should happen. Very philosophical and subtle difference 😄. But given that we advocate for the Given, When, Then, syntax it might make it more clear if we simply replaced `should` with `then`.
> 
> Example: `when added multiplied should return product` vs `when added multiplied then return product`
> 
> What do you think?

Good thought!

I don't think it is so important that we assert a specific phrasing, if some other word better communicates the expected outcome so that it is helpful for the future developer that reads it, then that should be fine to use. What I think is most important is that the pattern is there so that readers don't have to adjust the way they read the description. And more importantly don't have the parts either switching position or missing.

In the example with the Cat (even if that is probably not the best example) I think the parts are still clearly communicated.

But might still be good to update the examples so that they all use the same formatting, even if we don't want to lock down the phrasing.

PR #31, opened 2023-12-11
docs: Replace old table setup instructions in the auth module.

Isakdl:


docs/05-concepts/10-authentication/01-setup.mdx
```diff
@@ -33,36 +33,26 @@ modules:
 
 ### Initialize the auth database
 
-Start your database container from the server project.
+After adding the module to the server project, you need to initialize the database. First you have to create a new migration that includes the auth module tables. This is done by running the `serverpod create-migration` command line tool for the server project.
```
SandPod, line 36: Suggestion(Non-blocking): Mention that the command needs to be run from the server project directory.

The line ends with `for the server project.` maybe that could be replaced by `in the projects server directory`.

SandPod (approved):
👍

PR #32, opened 2023-12-12
docs: Improve ip configuration settings in the documentation.

Isakdl:
Closes: [#979](https://github.com/serverpod/serverpod/issues/979)


docs/05-concepts/01-working-with-endpoints.md
```diff
@@ -31,7 +32,20 @@ var client = Client('http://localhost:8080/')
   ..connectivityMonitor = FlutterConnectivityMonitor();
 ```
 
-If you run the app in an Android emulator, change the address to `10.0.2.2` as this is the IP address of the host machine.
+If you run the app in an Android emulator, change the address to `10.0.2.2` as this is the IP address of the host machine. To access the server from a different device on the same network (such as a physical phone) replace `localhost` with the local ip address. You can find the local ip by running `ifconfig` (Linux/MacOS) or `ipconfig` (Windows).
+
+Make sure to also update the `publicHost` in the development config to make sure the server always servers the client with the correct path to assets etc.
```
SandPod, line 37: NIT
```suggestion
Make sure to also update the `publicHost` in the development config to make sure the server always serves the client with the correct path to assets etc.
```

PR #33, opened 2023-12-12
docs: Expand custom serialization section with examples.

Isakdl:
- Remove duplicated content.
- Add detailed info on how custom serializable objects work.

Resolves documentation comment on: https://github.com/serverpod/serverpod/issues/858

I have opted to not write about the fact that you are able to write `project:src/path/to/file.dart` in the generator.yaml file. While this is possible should we encourage people to copy-paste code between two projects manually?

### DIFF
--- a/docs/05-concepts/02-protocol.md
+++ b/docs/05-concepts/02-protocol.md
@@ -1,6 +1,6 @@
 # Working with protocols
 
-Protocols are the YAML files used to define serializable classes in Serverpod. They are used to generate Dart code for the server and client, and, if a database table is defined, to generate database code for the server.
+Protocols are the YAML files used to define serializable classes in Serverpod. They are used to generate Dart code for the server and client, and, if a database table is defined, to generate database code for the server. The yaml-files are located in the `protocol` directory in your server project.
 
 The files are analyzed by the Serverpod CLI when generating the project and creating migrations.
 
@@ -15,6 +15,8 @@ fields:
   foundedDate: DateTime?
   employees: List<Employee>
 ```
+
+Supported types are `bool`, `int`, `double`, `String`, `DateTime`, `ByteData`, and other serializable classes. You can also use `List`s and `Map`s of the supported types, just make sure to specify the types. Null safety is supported. The keys of `Map` must be non-nullable `String`s. Once your classes are generated, you can use them as parameters or return types to endpoint methods.
 
 ### Limiting visibility of a generated class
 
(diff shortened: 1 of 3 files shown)


docs/05-concepts/02-protocol.md
```diff
@@ -1,6 +1,6 @@
 # Working with protocols
 
-Protocols are the YAML files used to define serializable classes in Serverpod. They are used to generate Dart code for the server and client, and, if a database table is defined, to generate database code for the server.
+Protocols are the YAML files used to define serializable classes in Serverpod. They are used to generate Dart code for the server and client, and, if a database table is defined, to generate database code for the server. The yaml-files are located in the `protocol` directory in your server project.
```
SandPod, line 3: Suggestion: Give full path.

```suggestion
Protocols are the YAML files used to define serializable classes in Serverpod. They are used to generate Dart code for the server and client, and, if a database table is defined, to generate database code for the server. The yaml-files are located in the `lib/src/protocol` directory in your server project.
```

SandPod: Should we also mention that they "should" or "need" to be located there to be picked up by the project?

docs/05-concepts/02-protocol.md
```diff
@@ -16,6 +16,8 @@ fields:
   employees: List<Employee>
 ```
 
+Supported types are `bool`, `int`, `double`, `String`, `DateTime`, `ByteData`, and other serializable classes. You can also use `List`s and `Map`s of the supported types, just make sure to specify the types. Null safety is supported. The keys of `Map` must be non-nullable `String`s. Once your classes are generated, you can use them as parameters or return types to endpoint methods.
```
SandPod, line 19: Question: are the types listed in many places? Maybe we should have then defined in a single place and reference it?

docs/05-concepts/03-serialization.md
```diff
@@ -1,67 +1,114 @@
-# Serialization
-Serverpod makes it easy to generate serializable classes that can be passed between server and client or used to communicate with the database.
+# Custom serialization
```
SandPod, line 1: Awesome! 🙌

SandPod (approved):
Great section about custom classes.

Couple of questions and suggestions for improvements.

docs/05-concepts/02-protocol.md
Isakdl: No it should be only here I believe, they where inside the serialisation page before.

PR #35, opened 2023-12-13
docs: Add documentation on future calls.

Isakdl:
Adds a new pages on the future calls feature.

Closes: https://github.com/serverpod/serverpod/issues/1305


docs/05-concepts/13-scheduling.md
```diff
@@ -0,0 +1,83 @@
...
+```dart
+await session.serverpod.futureCallAtTime(
+  'exampleFutureCall',
+  data,
+  DateTime(2025, 1, 1),
+);
+```
+
+:::note
+`data` is an object created from a class defined in one of your yaml files in the `protocol` folder, `data` may also be null if you don't need it.
```
SandPod, line 65: Suggestion: Mention that Data needs to match with the future call input type that was defined earlier.

SandPod (approved):
LGTM with my suggestion implemented to help users not run into runtime errors.

PR #38, opened 2023-12-18
refactor: Rename protocol to model.

Isakdl:
# Changes

All references to `protocol` and `entity` change to `model`. (Except, SerializebleEntity)


vlidholt (approved):
LGTM

PR #40, opened 2023-12-22
docs: Add how to opt out of migrations per table.

Isakdl:
Documentation for feature in PR: https://github.com/serverpod/serverpod/pull/1688

### DIFF
--- a/docs/05-concepts/06-database/11-migrations.md
+++ b/docs/05-concepts/06-database/11-migrations.md
@@ -5,6 +5,18 @@ A migration is a set of database operations (e.g. creating a table, adding a col
 
 If you ever get out of sync with the migration system, repair migrations can be used to bring the database schema up to date with the migration system. Repair migrations identify the differences between the two and create a unique migration that brings the live database schema in sync with a migration database schema.
 
+## Opt out of migrations
+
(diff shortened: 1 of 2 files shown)


SandPod (commented):
Looks good but please also add information about the extensions in this PR: https://github.com/serverpod/serverpod/pull/1688. That is, what happens when the flag is added/removed to existing tables.

Isakdl:
Done

docs/05-concepts/06-database/11-migrations.md
```diff
@@ -17,6 +17,8 @@ fields:
   name: String
 ```
 
+If you want to transition a manually managed table to then be managed by Serverpod you first need to set this flag to `true`. Then you have two options, either delete the old table you had created by yourself, and this will let Serverpod manage the schema from a clean state. However, this means that you would lose any data that was stored in the table. The second option is to make sure that the table schema matches the expected schema you have configured. This can be done by either manually making sure the schema aligns, or by creating a [repair migration](#creating-a-repair-migration) to get back into the correct state.
```
SandPod, line 20: Maybe the options are a bit more clear if you list them instead of describing them in the same block? Using a two point list or something?

SandPod (approved):
LGTM!

Maybe just double check the formatting again to make the two options stand out a bit from the text.

PR #41, opened 2023-12-22
expression improvement to be clear

Frank1016:
The original expression (>= 18 && 65 <=) is confusing for implementation. It should be improved to be clearer, such as  (t.age >= 18) & (t.age <= 65)


Isakdl:
Thank you!

PR #42, opened 2024-01-02
docs: Add relation to modules.

Isakdl:
Documentation for relation to modules feature in this PR: https://github.com/serverpod/serverpod/pull/1680

### DIFF
--- a/docs/05-concepts/06-database/03-relations/06-modules.md
+++ b/docs/05-concepts/06-database/03-relations/06-modules.md
@@ -1,0 +1,59 @@
+# Relations with modules
+
+Serverpod [modules](concepts/modules) usually come with predefined tables and data structures. Sometimes it can be useful to extend them with your data structures by creating a relation to the module tables. Relations to modules come with some restrictions since you do not own the definition of the table, you cannot change the table structure of a module table.
+In practice, this means you are never able to create a many-to-many relation as that would require a modification of the module model definition.
+
+## One to one
+
+Creating a relation to a module table is as easy as any other relation, simply reference the module class as the datatype. As for all one-to-one relations, we have to add a unique index to the relation field.
+
+```yaml
+class: User
+table: user
+fields:
+  userInfo: module:auth:UserInfo?, relation
+  age: int
+indexes:
+  user_info_id_unique_idx:
+    fields: userInfoId
+    unique: true
+```
+
+Or by referencing the table name if you only want to access the id.
+
+```yaml
+class: User
+table: user
+fields:
+  userInfoId: int, relation(parent=serverpod_user_info)
+  age: int
+indexes:
+  user_info_id_unique_idx:
+    fields: userInfoId
+    unique: true
+```
+
+## One to many
+
+A direct one-to-many relation is not supported as that would imply that the module table would have to be modified. But this limitation can be worked around by creating a "bridge" table and then creating a one-to-many relation to this new table instead!
+
+```yaml
+class: Employee
+table: employee
+fields:
+  userInfo: module:auth:UserInfo?, relation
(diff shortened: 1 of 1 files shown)


docs/05-concepts/06-database/03-relations/06-modules.md
```diff
@@ -0,0 +1,59 @@
+# Relations with modules
+
+Serverpod [modules](concepts/modules) usually come with predefined tables and data structures. Sometimes it can be useful to extend them with your data structures by creating a relation to the module tables. Relations to modules come with some restrictions since you do not own the definition of the table, you cannot change the table structure of a module table.
+In practice, this means you are never able to create a many-to-many relation as that would require a modification of the module model definition.
```
SandPod, line 4: Isn't it possible to create a many to many relation with a module using a junction table? That could be utilized without modifying the related table?

docs/05-concepts/06-database/03-relations/06-modules.md
```diff
@@ -0,0 +1,59 @@
...
+```yaml
+class: User
+table: user
+fields:
+  userInfo: module:auth:UserInfo?, relation
+  age: int
+indexes:
+  user_info_id_unique_idx:
+    fields: userInfoId
+    unique: true
```
SandPod, line 19: Wouldn't it be possible to skip the unique index if you would like to create a many to one relation? That is, multiple `Users` connected to one `UserInfo`?

SandPod: Suggest we also include that in the documentation.

docs/05-concepts/06-database/03-relations/06-modules.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+```yaml
+class: Company
+table: company
+fields:
+  employees: List<Employee>?, relation(name=company_employee)
+```
+
+As seen a bridge table just stores a reference to both related tables. We have added a unique index to the combination of the fields. This means we have restricted the data to have duplicated entries otherwise the same user could be included in the list twice.
```
SandPod, line 59:
```suggestion
As seen a bridge table just stores a reference to both related tables. A unique index is added to the combination of the fields. This prevents the user from being related to the same company multiple time.
```

SandPod: Is this actually want we want to do? Don't we want a unique index only on the userInfoId to prevent the same user from being connected to multiple companies? I think this index would allow the same user to be connected to many different companies and the same company to be connected to many different companies. Basically a many to many relation.

SandPod (commented):
Nice!

Some questions around the possibility of many to many relation and the exhaustiveness of configurability.

docs/05-concepts/06-database/03-relations/06-modules.md
Isakdl: But in this case you don't have a many-to-many with a UserInfo object, really. But you have it with your "shadow" UserInfo table.

Like you are not able to have an object relation on the UserInfo side either.

docs/05-concepts/06-database/03-relations/06-modules.md
Isakdl: But then it is no longer a bridge table but a one-to-many relation which you are not able to query with our query system.

docs/05-concepts/06-database/03-relations/06-modules.md
Isakdl: well in essence it is a many-to-many yeah. 

We restrict the employee to be unique for each company. But we do not restrict each user to only be employed at one company.

SandPod (approved):

PR #43, opened 2024-01-03
fix: Add warning about repair table removal.

SandPod:
Adds a warning that tables not part of the migration system will be removed by a repair migration and need to be managed manually.


Isakdl (approved):
LGTM

PR #44, opened 2024-01-03
Simplify 'Buld your first app tutorial' doc 

jakansha2001:


docs/03-tutorials/01-first-app.mdx
```diff
@@ -40,12 +40,17 @@ $ docker compose up --build --detach
...
+Make sure you have Docker running on your machine before executing these commands.
+
+:::
 
 ## Serialize objects
 
 Serverpod comes with a convenient way to create serializable objects with the help of code generation. These objects can easily be sent back and forth between the server and the client. This is managed by defining our objects in a YAML file which the code generator then parses and generates the necessary code for.
 
-To define the structure of our Note object, we will create a YAML file called `note.yaml`. Navigate to the `lib/src/model` folder in your Serverpod project (`notes_server`). Create the file and add the following content:
+To define the structure of our Note object, we will create a YAML file called `note.yaml` inside `lib/src/model` folder in your Serverpod project (`notes_server`). Add the following content in `note.yaml`:
```
SandPod, line 53:
```suggestion
To define the structure of our Note object, we will create a YAML file called `note.yaml` inside the `lib/src/model` directory in your Serverpod project (`notes_server`). Add the following content in `note.yaml`:
```

docs/03-tutorials/01-first-app.mdx
```diff
@@ -478,9 +542,15 @@ Widget build(BuildContext context) {
 }
 ```
 
-We trigger the note dialog when the action button is pressed and then save the note in the `onSaved` callback. To make the UI feel more responsive we add the changes to the notes list before calling `_createNote`. This way the note will be added to the list immediately and then updated when the server responds.
+Also import the dialog:
```
SandPod, line 545: Could the import be incorporated into the code snipped above instead since it is essentially part of the same step?

SandPod (approved):
Great improvements and clarifications for new users.

LGTM with two minor suggestions.

In the rest of our docs we prefer using directory instead of folder, but I noticed that is not true for this guide. So the first suggestion might not be relevant depending on what you think.

docs/03-tutorials/01-first-app.mdx
jakansha2001: Yes, we can but since we are only providing the code snippet of build method here, I thought of writing the import statement separately. This import statement will help the user get rid of the lint.

PR #48, opened 2024-01-10
docs: Fixes for release

vlidholt:


docs/05-concepts/06-database/06-filter.md
```diff
@@ -33,7 +33,7 @@ In the example we fetch all users with a name that is not Bob. If a non-`null` v
 
 ### Comparison operators
 
-Compare a column to a value, these operators are support for `int`, `double`, `Duration`, `DateTime` and `Enum`.
+Compare a column to a value, these operators are support for `int`, `double`, `Duration`, and `DateTime`. It's also available for `Enum`, if it is stored as an `integer`.
```
SandPod, line 36: This is a good catch, need to double check that we handle the difference here for enums.

Or did you check?

SandPod (approved):
Great fixes 🚀 Need to check the enum thing.

docs/05-concepts/06-database/06-filter.md
SandPod: We don't support any comparison operations for any Enums so it should be removed from this section.

PR #51, opened 2024-01-11
docs: :memo: update documentation

jakansha2001:
update get started doc and replace note class image to reflect model instead of protocol


vlidholt (approved):
LGTM

PR #52, opened 2024-01-15
fix: update migrations directory path.

SandPod:
Update the patch for the migration folder in the documentation.


jakansha2001:
LGTM :)

PR #59, opened 2024-01-19
fix: Use new order by syntax for documentation.

SandPod:
Resolve the last order by syntax issues with the documentation.


vlidholt (approved):
LGTM

PR #68, opened 2024-01-29
fix: model -> models

ewann:
A small fix, aim to become clear on contribution requirements before proposing a few more small fix PRs


SandPod (approved):
Thank you for making these fixes 🙏

PR #69, opened 2024-01-31
Update 02-models.md

Frank1016:
Correcting grammer errors

### DIFF
--- a/versioned_docs/version-1.2.0/05-concepts/06-database/02-models.md
+++ b/versioned_docs/version-1.2.0/05-concepts/06-database/02-models.md
@@ -29,7 +29,7 @@ fields:
 All fields are persisted by default and have an implicit `persist` set on each field.
 
 ### Data representation
-Storing a field with a primitive / core dart type will be handled as the their respective type. But if you use a complex type, such as another model, a `List` or a `Map` these will be stored as a `json` object in the database.
+Storing a field with a primitive / core dart type will be handled as its respective type. However, if you use a complex type, such as another model, a `List`, or a `Map`, these will be stored as a `json` object in the database.
 
 ```yaml
 class: Company


SandPod (commented):
Thank you for the fix!

We try to stay away from modifying the versions manually and instead regenerate the same version to amend any issues. 

Could you move these fixes to `docs/05-concepts/06-database/02-models.md` instead?

Frank1016:
> Thank you for the fix!
> 
> We try to stay away from modifying the versions manually and instead regenerate the same version to amend any issues.
> 
> Could you move these fixes to `docs/05-concepts/06-database/02-models.md` instead?

Actually, I don't know how to "move" the fix there. I went to that doc, and created another pull request there. If this is not the "move" you want, could you show me how to do it?

SandPod:
> Actually, I don't know how to "move" the fix there. I went to that doc, and created another pull request there. If this is not the "move" you want, could you show me how to do it?

What you did in the other PR is exactly what I was after, thank you.

I would like to ask one last thing from you in these PRs, could you please combine the changes into a single PR instead of having two of them? Since it is the same change we might as well add them together.

> In addition, can you show me the good way to propose fix suggestions when reading Serverpod docs?

The reason I wanted the change to be applied to the other file (as you have done in the other PR) is that the file that is modified in this PR is generated by the Docusaurus framework.

Once we are ready to release a new version of our documentation we generate a "version" that copies the contents of all the files in the `docs` directory into a newly created version in the `versioned_docs` directory.

Therefore, if we modify the versioned files in the `versioned_docs` directory, the changes would not be included in the future when we create new versions of our documentation.

We are using the framework Docusaurus to generate our documentation if you would like to read more.

> ...propose fix suggestions when reading Serverpod docs

I would say the same way that you did now! Please create a PR for anything that you would like to improve :)
Just make sure to target the file in the `docs` and possibly the versioned one if we want these changes to be visible directly.

Frank1016:
> > Actually, I don't know how to "move" the fix there. I went to that doc, and created another pull request there. If this is not the "move" you want, could you show me how to do it?
> 
> What you did in the other PR is exactly what I was after, thank you.
> 
> I would like to ask one last thing from you in these PRs, could you please combine the changes into a single PR instead of having two of them? Since it is the same change we might as well add them together.
> 
> > In addition, can you show me the good way to propose fix suggestions when reading Serverpod docs?
> 
> The reason I wanted the change to be applied to the other file (as you have done in the other PR) is that the file that is modified in this PR is generated by the Docusaurus framework.
> 
> Once we are ready to release a new version of our documentation we generate a "version" that copies the contents of all the files in the `docs` directory into a newly created version in the `versioned_docs` directory.
> 
> Therefore, if we modify the versioned files in the `versioned_docs` directory, the changes would not be included in the future when we create new versions of our documentation.
> 
> We are using the framework Docusaurus to generate our documentation if you would like to read more.
> 
> > ...propose fix suggestions when reading Serverpod docs
> 
> I would say the same way that you did now! Please create a PR for anything that you would like to improve :) Just make sure to target the file in the `docs` and possibly the versioned one if we want these changes to be visible directly.

Great, got it! Thanks a lot for your patience in explaining this to me!

PR #81, opened 2024-03-20
fix: Add migration guide for int -> bigint.

SandPod:
Add migration guide for int to bigint.

Documentation connected to this PR: https://github.com/serverpod/serverpod/pull/1194


Isakdl (approved):
LGTM

PR #82, opened 2024-03-20
Fix small mistake in 01-one-to-one.md

itsJoKr:
I believe this is a small mistake in docs. It's talking about an object field that does **not** need a parent keyword, and the example doesn't have a parent keyword.


itsJoKr:
Got it wrong.

PR #84, opened 2024-03-27
fix: document empty set behavior.

SandPod:
Add documentation for new empty set behavior with inSet and notInSet.


Isakdl (approved):
LGTM

PR #85, opened 2024-03-27
add explanation of .spy.yaml to "Working with models"

BenAuerDev:
Good morning @SandPod , 

As I was already looking into this, I figured I'd try and add it to the documentation :) 

let me know if something needs to be changed 

Ben

Closes: https://github.com/serverpod/serverpod/issues/2055

### DIFF
--- a/docs/05-concepts/02-models.md
+++ b/docs/05-concepts/02-models.md
@@ -1,6 +1,6 @@
 # Working with models
 
-Models are Yaml files used to define serializable classes in Serverpod. They are used to generate Dart code for the server and client, and, if a database table is defined, to generate database code for the server. Using regular `.yaml` files within `lib/src/models` is supported, but it is recommended to use `.spy.yaml` (.spy stands for "Server Pod Yaml") both within the models dir and when defining a model elsewhere within `lib/src`.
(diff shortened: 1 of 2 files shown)


docs/05-concepts/02-models.md
```diff
@@ -1,6 +1,6 @@
 # Working with models
 
-Models are the YAML files used to define serializable classes in Serverpod. They are used to generate Dart code for the server and client, and, if a database table is defined, to generate database code for the server. The yaml-files are located in the `lib/src/models` directory in your server project.
+Models are Yaml files used to define serializable classes in Serverpod. They are used to generate Dart code for the server and client, and, if a database table is defined, to generate database code for the server. Using regular `.yaml` files within `lib/src/models` is supported, but it is recommended to use `.spy.yaml` (.spy stands for "Server Pod Yaml") both within the models dir and when defining a model elsewhere within `lib/src`.
```
Isakdl, line 3: > and when defining a model elsewhere within `lib/src`

Suggestion: Remove this part as this is currently not a supported use-case. If you define the models elsewhere the code generator will not pick up the files.

Isakdl (changes requested):
Thank you for the PR :)!

I have requested one change for an invalid description. 

A minor addition that could be useful but not necessary is to add that if the .spy.yaml extension is used you get syntax highlighting with the Serverpod VSCode plugin.

BenAuerDev:
@Isakdl thanks for the quick feedback :) 

I removed the invalid description and added info about the VS Code Extension with a link to the marketplace. 

Let me know when it needs more tweaking :+1:

BenAuerDev:
also done :)

Isakdl (approved):
LGTM

PR #86, opened 2024-03-28
fix: Add section around how to set up custom authentication integration.

SandPod:
Adds a section explaining how to implement a custom authentication integration.

After review this addition should be added to the versioned docs so that it is available directly.

### DIFF
--- a/docs/05-concepts/10-authentication/05-custom-integration.md
+++ b/docs/05-concepts/10-authentication/05-custom-integration.md
@@ -1,0 +1,194 @@
+# Custom integration
+
+Serverpod is designed to make it as simple as possible to implement custom authentication integrations. The framework comes with an integrated auth token creation, validation, and communication system. With a simple setup, it is easy to generate tokens and include them in authenticated communication with the server.
+
+## Server setup
+
+After successfully authenticating a user, for example, through a username and password, an auth token can be created to preserve the authenticated user's permissions. This token is used to identify the user and facilitate endpoint authorization validation. When the user signs out, the token can be removed to prevent further access.
+
+### Create auth token
+
+To create an auth token, call the `signInUser` method in the `UserAuthentication` class, accessible through the `session.auth` field on the `session` object. 
+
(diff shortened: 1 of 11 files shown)


docs/05-concepts/10-authentication/05-custom-integration.md
```diff
@@ -0,0 +1,194 @@
...
+
+```dart
+var authToken = await session.auth.signInUser(....);
+var verifiableToken = '${authToken.userId}:${authToken.key}';
+```
+
+In the above example, the `verifiableToken` is created by concatenating the `userId` and `key` from the `authToken`. This token is then verifiable by the default token validation.
+
+#### Override token validation
+
```
vlidholt, line 45: Maybe clarify that this is typically not needed for building your own authentication method.

docs/05-concepts/10-authentication/05-custom-integration.md
```diff
@@ -0,0 +1,194 @@
...
+
+#### Custom token solution
+If a [custom auth token](#custom-tokens) solution has been implemented, auth token removal must be handled manually. The `signOutUser` method does not provide an interface to interact with other database tables.
+
+## Client setup
+Enabling authentication in the client is as simple as configuring a key manager and placing any token in it. If a key manager is configured, the client will automatically query the manager for a token and include it in communication with the server.
+
+### Configure key manager
+Key managers need to implement the `AuthenticationKeyManager` interface. The key manager is configured when creating the client by passing it as the named parameter `authenticationKeyManager`. If no key manager is configured, the client will not include tokens in requests to the server.
```
vlidholt, line 127: Mention that for most use cases, you can just use the `FlutterKeyManager` (or whichever we called it).

vlidholt (changes requested):
Overall super good, but it is a bit unclear what you _have_ to do to create your own authentication and what you _can_ do.

It would be nice to include a little bit on how to hook this up with the `UserInfo` class in the auth module too (as this will be the most common use case).

vlidholt (approved):
LGTM :)

PR #88, opened 2024-04-09
fix: update path to queries to configure database tables.

SandPod:
Update the path to the full sql definition in both 1.2 and next now that we have the migration system.


Isakdl (approved):
LGTM

PR #89, opened 2024-04-10
fix: Update return type for delete operations.

SandPod:
Documents the change of return type for the delete operations.

Code change PR: https://github.com/serverpod/serverpod/pull/2121


Isakdl (approved):
LGTM

PR #91, opened 2024-04-16
fix: add missing parameter, change incorrect member variable name

hampuslavin:
Fixes:
 - Missing `session` parameter in `updateRow` call
 - Unless I'm missing something, the member variable referenced in the docs, `authorizedUserId`, doesn't exist (anymore)?
   - I changed it to `authenticatedUserId` which is the existing one
   - Is this just a typo and should I change it in all versions of the docs? Or was this a breaking change at some point?

### DIFF
--- a/docs/05-concepts/06-database/05-crud.md
+++ b/docs/05-concepts/06-database/05-crud.md
@@ -98,7 +98,7 @@ To update a single row, use the `updateRow` method.
 ```dart
 var company = await Company.db.findById(session, companyId); // Fetched company has its id set 
 company.name = 'New name';
-var updatedCompany = await Company.db.updateRow(company);
+var updatedCompany = await Company.db.updateRow(session, company);
 ```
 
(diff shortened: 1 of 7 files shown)


SandPod (approved):
You are correct, authorizedUserId doesn't exist.
And I can't seem to find it in any of the older tags of the repo either so feel free to change on all version back to 1.0.0 (before that we don't have a tag so I'm not sure if it change names back then).

I'll approve and merge once the change is applied to 1.1.0 and 1.0.0

Thank you for the corrections 🙏

SandPod (changes requested):
Too quick on the trigger. Apply the `authorizedUserId` change on `1.1.0` and `1.0.0` and I'll approve and merge :)

hampuslavin:
@SandPod Done! ✅

SandPod (approved):
Thank you for this fix! 🙏

PR #93, opened 2024-04-20
Update 04-exceptions.md

akshaybengani:
I spent so much time figuring out why my client app was giving errors since I was not experienced with custom exceptions on the client side, I was like why is it not working then I looked on Google and found out this Exception name doesn't need brackets, Please merge this, many new developers like me can be stuck and then stress why my code is not working.

### DIFF
--- a/docs/05-concepts/04-exceptions.md
+++ b/docs/05-concepts/04-exceptions.md
@@ -41,7 +41,7 @@ In your app, catch the exception as you would catch any exception.
 try {
   client.example.doThingy();
 }
-on (MyException) catch(e) {
+on MyException catch(e) {
   print(e.message);
 }
 catch(e) {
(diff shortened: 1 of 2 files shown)


SandPod (changes requested):
Thank you for finding and fixing this issue!

I'm guessing Records might have change the debuggability of this a bit.

To preserve your change after we release new versions of our documentation it also needs to be applied to this file `docs/05-concepts/04-exceptions.md`. Please add it there as well and we will approve the change.

akshaybengani:
> Thank you for finding and fixing this issue!
> 
> I'm guessing Records might have change the debuggability of this a bit.
> 
> To preserve your change after we release new versions of our documentation it also needs to be applied to this file `docs/05-concepts/04-exceptions.md`. Please add it there as well and we will approve the change.

Done
Added the change in both the files
Thanks for quick resolution

PR #95, opened 2024-04-25
custom class serialization updates for the version 2.0

klkucaj:


SandPod (changes requested):
See comment.

docs/12-upgrading/01-upgrade-to-two.md
```diff
@@ -212,4 +212,36 @@ The authentication module's default value for allowing unsecure random number ge
 auth.AuthConfig.set(auth.AuthConfig(
   allowUnsecureRandom: true,
 ));
-```
\ No newline at end of file
+```
+
+## Updates to Custom Serialization in Serverpod 2.0
```
SandPod, line 217: Suggestion: This update changes all Model Serialization not just Custom Serialization since the parameter is not required when serializing Serverpod Models as well.

Please also describe that models can now be deserialized by calling `fromJson(...)` without the serialization manager parameter.

But I think it is good to mention how it effects the custom serialization as well. So these should probably be two separate sections.

SandPod (changes requested):

docs/12-upgrading/01-upgrade-to-two.md
```diff
@@ -212,4 +212,52 @@ The authentication module's default value for allowing unsecure random number ge
 auth.AuthConfig.set(auth.AuthConfig(
   allowUnsecureRandom: true,
 ));
+```
+
+## Updates to Serialization in Serverpod 2.0
+
+### General Changes to Model Serialization
+Serverpod 2.0 introduces a significant simplification in the serialization process of all models. Previously, the `fromJson` factory constructors required a `serializationManager` parameter to deserialize objects. This parameter is no longer needed, streamlining the deserialization process for all models, not just custom ones.
```
SandPod, line 220: The intention for the end of this line is repeated on line 236. Please remove it from here and only mention custom models down there.

SandPod (changes requested):

PR #96, opened 2024-04-25
Limit clause update for db examples

bfemmer:
Added a colon to the limit clause in the db examples to correct the typographical error.

Closes: https://github.com/serverpod/serverpod/issues/2158


SandPod (approved):
LGTM! 🎉

PR #99, opened 2024-05-04
remove leading "$" character from bash commands

### LINKED ISSUE #57: Remove the $ symbol in all md to get copy & paste more easy
Example:

Instead of this

```bash
$ cd mypod/mypod_server
$ docker compose up --build --detach
$ dart bin/main.dart --apply-migrations
```

use this

```bash
cd mypod/mypod_server
docker compose up --build --detach
dart bin/main.dart --apply-migrations
```

Then its is much easy to use the copy & paste function.

BenAuerDev:
This pr closes #57 

Changes:
1. Removes the leading "$" character so bash-commands can be copied and pasted without "$".

2. Split bash commands up so they can be copied individually as suggested in (https://github.com/serverpod/serverpod_docs/issues/57#issuecomment-2060936195)

Before:
cmd-before

After:
cmd-after

Also in "Getting started" I wasn't sure if I should repeat the "cd mypod..." command, I left it out for now. But if you want me to add it just let me know:

Before:
get-started-before

After:
get-started-after

### DIFF
--- a/docs/01-get-started.md
+++ b/docs/01-get-started.md
@@ -5,7 +5,7 @@ This page should give you an understanding of how a Serverpod project is structu
 Create a new project by running `serverpod create`.
 
 ```bash
-$ serverpod create mypod
+serverpod create mypod
 ```
 
 :::info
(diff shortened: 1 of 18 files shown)


BenAuerDev:
This became partially obsolete through #107

PR #102, opened 2024-05-15
Update documentation about password length

Katekko:
Today, I encountered some issues while troubleshooting the creation of a simple authentication flow. I utilized my own UI for the process and attempted to use `authController.createAccountRequest` to generate the account request, but unfortunately, it consistently failed.

Currently, there are no explicit error messages indicating the root cause of the failure; the method simply returns false. Consequently, I've first updated the documentation to reflect this challenge. Tomorrow, I plan to implement modifications to enhance error visibility during these steps.

I noticed that the sign-in method includes a logging mechanism for errors. Considering this, I'm considering creating a pull request (PR) to implement similar error logging in other methods within `server_pod_auth_email`.

### DIFF
--- a/versioned_docs/version-1.2.0/05-concepts/10-authentication/04-providers/01-email.md
+++ b/versioned_docs/version-1.2.0/05-concepts/10-authentication/04-providers/01-email.md
@@ -127,6 +127,10 @@ After the password has been reset you have to call the `signIn` method to log in
 
 Serverpod provides some additional configurable options to provide extra layers of security for stored password hashes.
 
+:::info
+By default, the minimum password length is set to 8 characters. If you wish to modify this requirement, you can utilize the properties within AuthConfig.
+:::
+
 ### Peppering
 
 For an additional layer of security, it is possible to configure a password hash pepper. A pepper is a server-side secret that is added, along with a unique salt, to a password before it is hashed and stored. The pepper makes it harder for an attacker to crack password hashes if they have only gained access to the database.


SandPod (changes requested):
Thanks for documenting this making it easer for all future developers of Serverpod to work with auth.

Please also apply this fix to `docs/05-concepts/10-authentication/04-providers/01-email.md` as it will otherwise be lost once we create a new release.

Also, we have increased debuggability of these things. If you enable debug logging in the server these reason for why the endpoints returns false should be revealed. These are available in Serverpod 1.2.7 and are implemented in this method: https://github.com/serverpod/serverpod/blob/13795a7bd4c0cc5a03101b6f378cb914673046dd/modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart#L374

Katekko:
Doc updated!

Great job on the log enhancements; debugging will now be much smoother. However, I'm considering whether it would be beneficial to throw errors to the user handler. What do you think? For instance, if the server encounters an error, wouldn't it be helpful to communicate custom exceptions to the UI? This way, users can be informed of specific issues, like attempting to register with a password below the minimum requirements, such as the failure to send an email.

Also, I'm curious about something: why wasn't a release/tag created on GitHub for version 1.2.7?

SandPod:
Awesome, thanks!

Unfortunately, we can't update the return type for that endpoint since there might be deployed clients depending on that behavior (But boy do we want to correct this!!). But if you look at the other authentication providers we give better error messages there.

We are looking at taking all the learnings from the current authentication module and create a new Auth2.0 module, but I can't provide a timeline for that unfortunately.

The 1.2.7 tag is a mistake on our end, we will add one ASAP! Thanks for letting us know!

SandPod (approved):
🎉

PR #103, opened 2024-05-15
fix: Update docs around removed fields.

SandPod:
Documents the breaking changes introduced here: https://github.com/serverpod/serverpod/pull/2249


Isakdl (approved):
LGTM

PR #104, opened 2024-05-15
Fix: findRow -> findFirstRow

kieran-myopnotes:
Updated 05-crud.md from old syntax to new


Isakdl:
Looking good, but could you also add this into the files inside: `versioned_docs/version-1.2.0` so it is updated for 1.2 as well as the `next` version.

kieran-myopnotes:
I have also updated the 1.2 versions now as well

PR #105, opened 2024-05-17
Document auth config

BenAuerDev:
This PR tries to close serverpod/serverpod#2245 and also holds the documentation part of serverpod/serverpod#2250.

Please let me know where there is room for improvement :) 

docss


Isakdl (changes requested):
Great contribution! :) Thank you!

I have two smaller change requests for accuracy!

Edit: Please also add the change to 1.2 documentation as well!

PR #106, opened 2024-05-17
fix: Add redirect for moved documentation pages.

SandPod:
Adds a client side redirect for moved documentation pages.

This plugin generates a static html for the pages that have been removed and redirects on the client side to the new url.


Isakdl (approved):
LGTM

PR #107, opened 2024-05-17
Reformat docs

### LINKED ISSUE #100: Consider using consistent formatting for .md files
I have a markdown linter (DavidAnson.vscode-markdownlint) activated and when opening a .md file in the docs all the yellow can be quite intimidating, plus I can't use format-on-save.

My proposal: Format existing .md files according to markdownlint and require correct formatting like done for dart code.

unformatted:
unformatted

formatted: 
formatted

PS: Of course I volunteer for the reformatting :)

BenAuerDev:
This PR tries to close #100 

## Description
1. All .md files in the "next" documentation where fixed according to markdownlint standard with a few rules ignored (explanation on the bottom) 

2. A CI job was added to test-build.yaml testing for any markdownlint violations (that were not ignored)

**set up ignore** in one commit 
**auto fixes** in one large commit
**manual fixes** individual per fix

A visible change was introduced as I had to rebuild a table without using inline html, which made it full-width (which i find nicer but we can also ignore that rule)

<div float="left" width="100">
  
   
</div>

## Ignored rules 

I also added a `.markdownlint.json` (can also be done in yaml if preferred) which ignores a few rules. 

These rules introduce changes that I would like to have discussed before automatically applying it. 

`MD001:` Enforces consistent heading levels. For example, a level 2 heading should not be followed directly by a level 4 heading. 
Comment: This only happens in "docs/05-concepts/06-database/01-connection.md"

`MD013`: Enforces a maximum line length. By default, it triggers a warning for lines longer than 80 characters. 
Comment: This I put in there by default because there are lines that are more than 80 but I think it's fine.

`MD014`: Warns when dollar signs are not used in shell code blocks.
Comment: Not sure if you want to keep those

`MD024`: Prevents multiple headings with the same content.
Comment: `docs/05-concepts/06-database/07-relation-queries.md` and `docs/05-concepts/06-database/09-pagination.md` both have `### Tips` and `### Nested Includes` (which I think is fine)


.markdownlint.json
```diff
@@ -0,0 +1,8 @@
+{
+    "MD001": false,
+    "MD013": false,
+    "MD014": false,
+    "MD024": false,
+    "MD025": false,
+    "MD029": false
+}
```
Isakdl, line 8: Suggestion: Convert to yaml, I think it is less verbose to edit in general.

Isakdl (changes requested):
@BenAuerDev MVP today! :) 

Thank you very much for this PR!

Aside from the minor comment, I think we should enable `MD014` if this enforces `$` in shell command blocks? Because we do want that in the docs.

I also think `MD025` makes sense to enforce and if there is only one page that is not following it we should simply change that to a ##.

I have no problem enforcing `MD029` I actually think the document becomes more readable so probably a good idea to keep that.

But good work! 🚀

BenAuerDev:
> @BenAuerDev MVP today! :)
> 
> Thank you very much for this PR!
> 
> Aside from the minor comment, I think we should enable `MD014` if this enforces `$` in shell command blocks? Because we do want that in the docs.
> 
> I also think `MD025` makes sense to enforce and if there is only one page that is not following it we should simply change that to a ##.
> 
> I have no problem enforcing `MD029` I actually think the document becomes more readable so probably a good idea to keep that.
> 
> But good work! 🚀

Thanks :) Glad you like it

`M014` works the other way around, it would remove the leading "$" character from bash commands :) 

But I'll remove MD025 and make it a ## :)

BenAuerDev:
ups the last commit should be named DM025

Question: Should we mention in the contribution docs somewhere that code should be formatted

Isakdl:
> Question: Should we mention in the contribution docs somewhere that code should be formatted

Good point, on that do you have a formatter that can easily be installed? I mean CLI tool, I think that would be the best.
And yes probably a good idea to add in the readme of this project.

And the rule 29 if that enforces this style:
```
1. List opening
    Related Text....
2. List continuation
    Related Text....
```

I think we should enable that rule too.

BenAuerDev:
yeah super easy 
```bash
npm install -g markdownlint-cli
```
to install it globally on the machine. 

but there is also a [vs-code-extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) with solid auto-format on save :) 

should I add both the command and a link to the extension to the readme? :) 

Ok I'll enable MD029 and apply the fixes to it :)

Isakdl:
> should I add both the command and a link to the extension to the readme? :)

Yes please do so! :)

BenAuerDev:
@Isakdl done :D

Isakdl:
Last small request, could you add a lint command into package.json?

BenAuerDev:
absolutely :) already done :D

Isakdl (approved):
LGTM

PR #109, opened 2024-05-17
ci: Copy CNAME on docs deploy.

SandPod:
When the deploy repo was initialized there probably was a manual step that involved copying over the CNAME file to the docs repo.

Since we now clear out the docs repo between deploys the CNAME file was accidentally removed.

We now always copy over the CNAME file in the deploy pipeline.


Isakdl (approved):
LGTM

PR #110, opened 2024-05-20
Rewrote outdated raw sql queries docs to match latest parameter binding update

Aurangseb:
This document was quite outdated and included functions that are not even there any more. Pretty much a rewrite from scratch.

### DIFF
--- a/docs/05-concepts/06-database/10-raw-access.md
+++ b/docs/05-concepts/06-database/10-raw-access.md
@@ -75,6 +75,7 @@ var result = await db.unsafeQuery(
 );
 ```
 
-> [!CAUTION]
-> Always sanitize your input when using raw query methods. For the `unsafeQuery` and `unsafeExecute` methods, use query parameters to prevent SQL injection. Avoid using `unsafeSimpleQuery` and `unsafeSimpleExecute` unless the simple query protocol is strictly required.
+:::danger
+Always sanitize your input when using raw query methods. For the `unsafeQuery` and `unsafeExecute` methods, use query parameters to prevent SQL injection. Avoid using `unsafeSimpleQuery` and `unsafeSimpleExecute` unless the simple query protocol is strictly required.
+:::
 
--- a/docs/05-concepts/06-database/10-raw-access.md
+++ b/docs/05-concepts/06-database/10-raw-access.md
@@ -1,45 +1,80 @@
-# Raw access
+# Raw Access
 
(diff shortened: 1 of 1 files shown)


Aurangseb:
Docs are related to: https://github.com/serverpod/serverpod/pull/2156

docs/05-concepts/06-database/10-raw-access.md
```diff
@@ -1,45 +1,80 @@
...
+DatabaseResult result = await session.db.unsafeQuery(
+  r'SELECT * FROM mytable WHERE id = @id', 
+  parameters: QueryParameters.named({'id': 1}),
 );
 ```
 
-Use the `unsafeQueryMappedResults` method when making complex queries with joins. The method returns a `List<Map<String, Map<String, dynamic>>>`, where the `List` contains an entry for each result row. The result row is a `Map` where the key is the table from which the data has been retrieved and the value is another `Map` of that tables columns. The column `Map` inside the table `Map` has column names as keys and the contents of the colum as values.
+## `unsafeExecute`
+
+Executes a single SQL query without returning any results. Use this for statements that modify data, such as `INSERT`, `UPDATE`, or `DELETE`. Returns the number of rows affected.
```
Isakdl, line 18: Suggestion: Change `without returning any results` as the method returns the number of effected rows.

docs/05-concepts/06-database/10-raw-access.md
```diff
@@ -1,45 +1,80 @@
...
+// Positional parameters
+var result = await db.unsafeQuery(
+  r'SELECT id FROM apparel WHERE color = $1 AND size = $2',
+  QueryParameters.positional(['green', 'XL']),
 );
 ```
 
-in the above example, the `result` variable will contain the number of rows that were deleted.
+> [!CAUTION]
+> Always sanitize your input when using raw query methods. For the `unsafeQuery` and `unsafeExecute` methods, use query parameters to prevent SQL injection. Avoid using `unsafeSimpleQuery` and `unsafeSimpleExecute` unless the simple query protocol is strictly required.
```
Isakdl, line 79: Suggestion:

Keep the
```md
:::danger
content
:::
```

As it renders a nice visual box

Screenshot 2024-05-20 at 13 12 22

Isakdl (changes requested):
Very nice! Thank you for the PR!

I have two small requested changes but other than that looking good! 👍

Isakdl (approved):
LGTM

PR #122, opened 2024-06-06
mini doesn't require DB

KyleFin:


versioned_docs/version-2.0.0/06-deployments/05-general.md
```diff
@@ -4,7 +4,7 @@ You can host Serverpod anywhere, running Dart directly or through a Docker conta
 
 ## Required services
 
-Serverpod will not run without a link to a Postgres database with the correct tables added. Serverpod can also optionally use Redis. You enable Redis in your configuration files.
+Serverpod will not run without a link to a Postgres database with the correct tables added (unless you're running Serverpod mini). Serverpod can also optionally use Redis. You enable Redis in your configuration files.
```
KyleFin, line 7: Is this understanding correct? (Based on https://medium.com/serverpod/getting-started-with-serverpod-mini-58c71ffefa54)

SandPod: Yes, this is correct! 👍

SandPod (changes requested):
Another good catch, please also apply it to the non versioned docs, located here `docs/06-deployments/05-general.md`, so that the change will be preserved once we release a new version of our documentation.

SandPod (approved):
🚀

PR #125, opened 2024-06-20
adding docs for new default keywords

klkucaj:


docs/05-concepts/02-models.md
```diff
@@ -159,3 +159,51 @@ extension MyExtension on MyClass {
...
+
+## Default Values
+
+Serverpod supports defining default values for fields in your models. These defaults can be specified using the following keywords:
+
+#### Keywords
+
+- **default**: Sets the default value for both the model and the database.
+- **defaultModel**: Sets the default value for the model side.
+- **defaultDatabase**: Sets the default value for the database side.
```
SandPod, line 171: Suggestion: Add these keywords to the table above that lists all keywords supported for a model.

docs/05-concepts/02-models.md
```diff
@@ -159,3 +159,51 @@ extension MyExtension on MyClass {
 |[**fields (index)**](database/indexing)|List the fields to create the indexes on.                                                |✅|||
 |[**type (index)**](database/indexing)|The type of index to create.                                                               |✅|||
 |[**unique**](database/indexing)|Boolean flag to make the entries unique in the database.                                         |✅|||
+
+## Default Values
```
SandPod, line 163: Suggestion: Add a simple example demonstrating how to configure a default value to the introduction.

This is most likely what most people are looking for.

docs/05-concepts/02-models.md
SandPod: If these are listed in the list above then we don't need to repeat that information here and this section can be removed.

docs/05-concepts/02-models.md
```diff
@@ -159,3 +159,51 @@ extension MyExtension on MyClass {
...
+
+#### Keywords
+
+- **default**: Sets the default value for both the model and the database.
+- **defaultModel**: Sets the default value for the model side.
+- **defaultDatabase**: Sets the default value for the database side.
+
+#### Supported Default Values
+
+- **DateTime**: Supports "now" (current date and time) and DateTime formatted strings (only in UTC).
```
SandPod, line 175: Suggestion: Make `DateTime` a header and list the supported ways of formatting it more clearly.

Maybe add an example for each so that it is easy to understand how it should be formatted.

docs/05-concepts/02-models.md
```diff
@@ -159,3 +159,51 @@ extension MyExtension on MyClass {
...
+  ### Sets a specific DateTime formatted string in UTC as the default value for the database side.
+  ### If no value is set, this default will be used when the field is persisted in the database.
+  ### Fields with 'defaultDatabase' must be nullable.
+  dateTimeDatabase: DateTime?, defaultDatabase=2024-05-34T22:00:00.000Z
+
+  ### Sets different default values for the model and the database sides.
+  dateTimeModelAndDatabase: DateTime, defaultModel=2024-05-01T22:00:00.000Z, defaultDatabase=2024-05-10T22:00:00.000Z
+```
+
+- **Other Types**: Currently, default values are only supported for the DateTime field type.
```
SandPod, line 204: Suggestion: Remove this since we are listing what is supported.

docs/05-concepts/02-models.md
```diff
@@ -159,3 +159,51 @@ extension MyExtension on MyClass {
...
+
+### Restrictions
+
+- **relation** keyword cannot be used with defaults.
+- **!persist** keyword cannot be used when **defaultDatabase** is specified.
```
SandPod, line 209: Question: This can be included when we describe the keyword right? I think that is the case for the other keywords. Then this section can be removed completely.

SandPod (commented):
Couple of suggestions for improvements.

SandPod (approved):

PR #132, opened 2024-08-11
docs: added a new section to getting started for iOS simulator

Code-Remote:
I wanted to get started with ServerPod but it failed on my chrome browser, since i had ran it before and it worked I was too lazy to debug it so I thought I'll try it out on my iPhone simulator. There it also crashed, then I tried Android emulator, there it worked. Apparently the SocketException on iOS was due to the fact that an iPhone simulator uses localhost for itself. After changing local host to the ip address of my own machine it worked.

### DIFF
--- a/docs/01-get-started.md
+++ b/docs/01-get-started.md
@@ -99,12 +99,9 @@ The flag `-d chrome` runs the app in Chrome, for other run options please see th
 
 :::info
 
-### iOS Simulator
-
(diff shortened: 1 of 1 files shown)


SandPod:
Hey @Code-Remote!

Thank you for the contribution to the documentation! 
We will take a look at this shortly! :)

SandPod (approved):
Nice work.

Thanks for the addition and helping other developers out in the future 🚀

PR #136, opened 2024-08-15
docs: Adding docs for default values on UuidValue model field type

klkucaj:


docs/06-concepts/02-models.md
```diff
@@ -226,9 +226,31 @@ Supports any string value, e.g., `"This is a string"` or `'This is a string'`.
...
+
+2. **Random UUID**:
+   - You can use `"random"` to specify a random UUID. On the Dart side, this will be internally handled by `Uuid().v4obj()` to generate a UUID at runtime. On the database side, the `gen_random_uuid()` function is used to generate the UUID.
+
+   Example:
+
+   ```yaml
+   uuidDefaultRandom: UuidValue, default=random, defaultModel=random, defaultPersist=random
+   ```
+
```
SandPod, line 253: Question: Could we improve the formatting here? It looks a bit funky in and it is a bit unclear that these are key values. Maybe a table could be used instead? Same for DateTime `now` and string value.

docs/06-concepts/02-models.md
```diff
@@ -226,9 +226,31 @@ Supports any string value, e.g., `"This is a string"` or `'This is a string'`.
...
-stringDefault: String, default='This is a string', defaultModel="This is a string", defaultDatabase="This is a string"
+stringDefault: String, default='This is a string', defaultModel="This is a string", defaultPersist="This is a string"
 ```
 
+### UuidValue
+
+Supports two types of default values:
+
+1. **UUID String Values**:
+   - A valid UUID string can be specified, e.g., `'550e8400-e29b-41d4-a716-446655440000'`.
```
SandPod, line 237: Question: Only v4 UUIDS are expected right? Maybe we should have a note on that.

docs/06-concepts/02-models.md
```diff
@@ -243,18 +265,21 @@ fields:
...
 
   ### Sets the default value for a double field.
-  doubleDefault: double, default=10.5, defaultDatabase=20.5
+  doubleDefault: double, default=10.5, defaultPersist=20.5
 
   ### Sets the default value for a string field.
   stringDefault: String, default="This is a string", defaultModel="This is a string"
+
+  ### Sets the default value for a UUID field.
+  uuidDefault: UuidValue, default=random, defaultModel=random, defaultPersist=random
```
SandPod, line 280: Question: Does this add any value to the point proven in this section? Otherwise it could probably be removed to keep the example as simple as possible.

SandPod (commented):
Nice work! Couple of questions regarding formatting.

docs/06-concepts/02-models.md
klkucaj: I just made the UUID part more clear, let me know if its still not good?!

SandPod (commented):
LGTM!

But lets merge this once the real PR is in also.

docs/06-concepts/02-models.md
```diff
@@ -161,72 +161,98 @@ extension MyExtension on MyClass {
...
+| Type                       | Keyword | Description |
+|----------------------------|---------|-------------|
+| **Boolean Values**   | `true` or `false`  | Sets the field to a boolean value, either `true` or `false`. |
 
-Example:
+**Example:**
 
 ```yaml
-boolDefault: bool, default=true, defaultModel=false, defaultDatabase=true
+boolDefault: bool, default=true, defaultModel=false, defaultPersist=true
```
Isakdl, line 197: Suggestion: Move `default=true, defaultModel=false, defaultPersist=true`these variables to its own section in the top and explain the difference between them and how they interact.

Then for each type only use the `default` keyword as the example.

docs/06-concepts/02-models.md
```diff
@@ -161,72 +161,98 @@ extension MyExtension on MyClass {
 |[**unique**](database/indexing)|Boolean flag to make the entries unique in the database.                                         |✅|||
 |[**default**](#default-values)|Sets the default value for both the model and the database. This keyword cannot be used with **relation**.                                                     |✅|||
 |[**defaultModel**](#default-values)|Sets the default value for the model side. This keyword cannot be used with **relation**.                                                   |✅|||
-|[**defaultDatabase**](#default-values)|Sets the default value for the database side.  This keyword cannot be used with **relation** and **!persist**.                                |✅|||
+|[**defaultPersist**](#default-values)|Sets the default value for the database side.  This keyword cannot be used with **relation** and **!persist**.                                |✅|||
 
 ## Default Values
```
Isakdl, line 166: Suggestion: Move this section above the `Keywords` section. The keywords should come last!

docs/06-concepts/02-models.md
```diff
@@ -243,18 +269,18 @@ fields:
...
+  doubleDefault: double, default=10.5, defaultPersist=20.5
 
   ### Sets the default value for a string field.
   stringDefault: String, default="This is a string", defaultModel="This is a string"
 ```
 
 ### Coming Soon
 
-- **UuidValue**: Support for default values for `UuidValue` field type is coming soon.
+- **Duration**: Support for default values for the `Duration` field type is coming soon.
```
Isakdl, line 286: Suggestion: Remove any coming soon docs. I think it is generally a bad idea to add "coming soon" things in our docs. If we for some reason do not implement a feature it looks pretty bad with a bunch of coming soon things spread around that are not being implemented. 

Instead only describe functional features that can be used.

docs/06-concepts/02-models.md
```diff
@@ -161,72 +161,98 @@ extension MyExtension on MyClass {
 |[**unique**](database/indexing)|Boolean flag to make the entries unique in the database.                                         |✅|||
 |[**default**](#default-values)|Sets the default value for both the model and the database. This keyword cannot be used with **relation**.                                                     |✅|||
 |[**defaultModel**](#default-values)|Sets the default value for the model side. This keyword cannot be used with **relation**.                                                   |✅|||
```
Isakdl, line 163: Question: I suppose this has not been added to exceptions, but doesn't it make sense to also add this feature for the exception type? It should be rather trivial to add too.

Isakdl (changes requested):
Had some thoughts on some smaller docs changes! :)

docs/06-concepts/02-models.md
klkucaj: I think this is better to it in another pr.

Isakdl (approved):
LGTM! 🚀

PR #143, opened 2024-08-15
chore: add groups for docusaurus and react dependencies.

SandPod:
Creates two new dependency groups that ensure that all react dependencies are bumped together and all docusarus dependencies are bumped together.


hampuslavin (approved):
🌟

PR #149, opened 2024-08-19
fix: Add documentation for how to update a specific column.

Isakdl:
Adds documentation for updates targeting a single column.


SandPod (approved):
LGTM! 👍

PR #150, opened 2024-08-21
docs: Adding duration default value docs

klkucaj:


docs/06-concepts/02-models.md
```diff
@@ -274,6 +288,9 @@ fields:
 
   ### Sets the default value for a string field.
   stringDefault: String, default="This is a string", defaultModel="This is a string"
+
+  ### Sets the default value for a duration field.
+  durationDefault: Duration, default=1d 2h 10min 30s 100ms
```
SandPod, line 293: Question: Does this add any value?

docs/06-concepts/02-models.md
```diff
@@ -250,6 +250,20 @@ uuidDefaultRandom: UuidValue, default=random
 uuidDefaultUuid: UuidValue, default='550e8400-e29b-41d4-a716-446655440000'
 ```
 
+### Duration
+
+#### Supported Default Values
+
+| Type                          | Keyword  | Description |
+|-------------------------------|----------|-------------|
```
SandPod, line 258: Suggestion: Please describe what the "keys" mean.

That "d" is day and "min" is minutes etc.

SandPod (commented):
Two comments.

Could you also arrange the default value types in alphabetical order? Right now the order feels a bit Random?

docs/06-concepts/02-models.md
klkucaj: Done!

docs/06-concepts/02-models.md
klkucaj: nop, removed!

SandPod (approved):
LGTM 👍
Only thing I think about now is the header level in the menu, where Support types are mixed with other headers: 
Screenshot 2024-08-22 at 08 27 50

Can we do something about this?

klkucaj:
@SandPod check my last commit about the menu.

klkucaj:
> LGTM 👍 Only thing I think about now is the header level in the menu, where Support types are mixed with other headers: Screenshot 2024-08-22 at 08 27 50
> 
> Can we do something about this?

@SandPod check my last commit about the menu.

PR #153, opened 2024-08-23
docs: add all needed host aliases

hampuslavin:
- Adds additional aliases that are needed to run integration tests locally


SandPod (approved):
LGTM! 👍

PR #156, opened 2024-08-27
docs: Adding docs for default values on Exception fields

klkucaj:


docs/06-concepts/04-exceptions.md
```diff
@@ -50,3 +50,35 @@ catch(e) {
   print('Something else went wrong.');
 }
 ```
+
+### Default Values in Exceptions
+
+Serverpod supports the use of default values for fields in exceptions, just like it does for models. This feature allows you to define default values for any supported field type directly within your exception definitions.
```
SandPod, line 56: Suggestion: Iterate this with GPT.

This sentence feels a bit strange; how about running it through ChatGPT a couple of times to improve it?

docs/06-concepts/04-exceptions.md
```diff
@@ -50,3 +50,35 @@ catch(e) {
...
+```
+
+In the example above:
+
+- The `message` field will default to `"An error occurred"` if not provided.
+- The `errorCode` field will default to `1001`.
+
+#### Important Note
+
+Since exceptions are never persisted to the database, we do not support `defaultPersist` for exception fields. This means that only `default` and `defaultModel` are relevant in this context.
```
SandPod, line 78:
```suggestion
Since exceptions are never persisted in the database, the `defaultPersist` keyword is not supported.
```

SandPod: Maybe even add it as a `::note` or `::info` depending on what looks best.

docs/06-concepts/04-exceptions.md
```diff
@@ -50,3 +50,35 @@ catch(e) {
...
+
+However, it’s important to note that `defaultModel` will override `default` if both are specified. Since exceptions only exist in the model (Dart code) and are not persisted, using `defaultModel` in combination with `default` is redundant. The `defaultModel` value will always take precedence over `default`, making it unnecessary to use both in conjunction.
```
SandPod, line 80: Suggestion: Shorten this down a bit.

I think it can probably be explained in a single sentence.

docs/06-concepts/04-exceptions.md
```diff
@@ -50,3 +50,35 @@ catch(e) {
...
+
+#### Learn More
+
+For a complete explanation of how default values work, including the supported types and the behavior of the `default` and `defaultModel` keywords, please refer to the [Default Values](models#default-values) section in the [Working with Models](models) documentation.
```
SandPod, line 84: Suggestion: Use ChatGPT to improve this.

I think this could also be improved a bit using ChatGPT.
This could also focus a bit more on the supported types and how they are specified.

docs/06-concepts/04-exceptions.md
SandPod: We could probably already refer to the models section in the introduction but then specify that only `default` and `defaultModel` are supported.

SandPod (commented):
Think this needs another round of polish, probably shorten it down a bit.

docs/06-concepts/04-exceptions.md
klkucaj: i went with ::info, its better

SandPod (approved):
LGTM! 🚀

PR #158, opened 2024-08-29
fix: Add documentation for authentication in streaming methods.

SandPod:


docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -42,6 +46,36 @@ List<String> scopes = extractScopes(token);
...
+### Handling revoked authentication 
+
+When a user's authentication is revoked, the server must be promptly notified to maintain secure access control. The `authenticationRevoked` event plays a crucial role in closing method streams that the user is no longer authorized to access. To trigger this event, invoke the `session.messages.authenticationRevoked` method.
+
+```dart
+await session.messages.authenticationRevoked(
+  userId,
+  RevokedAuthenticationScope(
+    scopes: removedScopes,
+  ),
```
hampuslavin, line 58: Question (for my understanding): From an API point of view, why are we sending the diff (`removedScopes`) in the event instead of the actual scopes that the user has? (For some reason I didn't think about this when I reviewed the actual PR) e.g.
```
await session.messages.authenticationRevoked(
  userId,
  RevokedAuthenticationScope(
    newScopes: newScopes,
  )
```
I would imagine in most use cases, it would be more straight forward to pass in the current state of things?

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -42,6 +46,36 @@ List<String> scopes = extractScopes(token);
...
+);
+```
+
+##### Parameters
+
+- `userId` - The user id for the concerned user. This value should match the `userId` field in the `AuthenticationInfo` object.
```
hampuslavin, line 64: If opting for maximum clarity, I would suggest to be more explicit
```suggestion
- `userId` - The user id of the user to be revoked. This value should match the `userId` field in the `AuthenticationInfo` object.
```
Although I would challenge this to be written as this for brevity:
```suggestion
- `userId` - The user id belonging to the `AuthenticationInfo` object to be revoked.
```

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -42,6 +46,36 @@ List<String> scopes = extractScopes(token);
...
+- `RevokedAuthentication` - The extent of the authentication revocation.
```
hampuslavin, line 65: Suggestion 1: The previous bullet was the parameter _name_, whereas this is the parameter _type_. I would suggest to change this to the parameter name as well (`message` I think?).

Suggestion 2: Be more strictly technical (more like code doc comments):
```suggestion
- `message` - The revoked authentication event message. See below for the different types of messages.
```

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -42,6 +46,36 @@ List<String> scopes = extractScopes(token);
 Set<Scope> userScopes = scopes.map((scope) => Scope(scope)).toSet();
 ```
 
+### Handling revoked authentication 
+
+When a user's authentication is revoked, the server must be promptly notified to maintain secure access control. The `authenticationRevoked` event plays a crucial role in closing method streams that the user is no longer authorized to access. To trigger this event, invoke the `session.messages.authenticationRevoked` method.
```
hampuslavin, line 51: Suggestion: Remove adverbs/adjectives and "leaking implementation" details

Two concrete examples just to explain what I'm getting at:
- "promptly" is superfluous info, the user either must or mustn't do it
- Wether the event "plays a crucial role" or not is beside the point (implementation detail), the user just needs to know what event needs to be raised
- "maintain secure access control" is vague, while something explicit like "the server will respect the new changes" is unambiguous 

The following text contains the same info with these things removed:

> When a user's authentication is revoked, the server must be notified to respect the changes (e.g. to close method streams). Invoke the `session.messages.authenticationRevoked` method and raise the appropriate event to notify the server.

Let me know what you think, as it's also a matter of writing style. Personally I prefer to go through technical documentation with as crisp explanations as possible because that makes it easier to find what I'm looking for when searching/skimming.

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -42,6 +46,36 @@ List<String> scopes = extractScopes(token);
...
+);
+```
+
+##### Parameters
+
+- `userId` - The user id for the concerned user. This value should match the `userId` field in the `AuthenticationInfo` object.
+- `RevokedAuthentication` - The extent of the authentication revocation. 
+
+
+#### Revoked authentication types
```
hampuslavin, line 68: Question: Isn't this an _event message_ rather than a _type_? In that case the header should reflect that
```suggestion
#### Revoked authentication messages
```

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -42,6 +46,36 @@ List<String> scopes = extractScopes(token);
...
+There are three types of `RevokedAuthentication` objects that are used to specify the extent of the authentication revocation:
+
```
hampuslavin, line 70:
```suggestion
There are three types of `RevokedAuthentication` messages that are used to specify the extent of the authentication revocation:

```

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -42,6 +46,36 @@ List<String> scopes = extractScopes(token);
...
+| Type | Description |
+|-----------|-------------|
+| `RevokedAuthenticationUser` | Informs that all authentication is revoked for a user. |
+| `RevokedAuthenticationAuthId` | Informs that a single authentication id is revoked for the user. This should match the `authId` field in the `AuthenticationInfo` object. |
+| `RevokedAuthenticationScope` | Informs that a list of scopes have been revoked for the user. |
+
+Each type provides a tailored approach to revoke authentication based on different needs, ensuring that access is precisely controlled and managed.
```
hampuslavin, line 77: Suggestion 1: replace "Type" with "Message" or "Message type"

Suggestion 2: remove "Informs that". It doesn't add any new information. Since we're telling the user to use these events, it is implicit that they will inform the server.

hampuslavin (commented):
Super nit-picky review here 🙋 😃 Not saying I'm right by any means but I just want to make sure I understand and are aligned on how we write docs.

In these types of technical docs (i.e. NOT a tutorial), I tend to value _brevity_ and _explicitness_ highest.

docs/06-concepts/11-authentication/05-custom-overrides.md
SandPod: No, we send the revoked scopes, so the example is correct. To make this even clearer, removed scopes should probably be called `revokedScoped` in the example.

The purpose of this is to make it as easy and declarative as possible to post only what is revoked.

docs/06-concepts/11-authentication/05-custom-overrides.md
SandPod: Great suggestion! I will use the second one; it seems like a good simplification!

docs/06-concepts/11-authentication/05-custom-overrides.md
SandPod: Great suggestion! Simplification is everything. Will use this

docs/06-concepts/11-authentication/05-custom-overrides.md
SandPod: Updated the example to reflect this as well.

docs/06-concepts/11-authentication/05-custom-overrides.md
hampuslavin: Thanks!

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -42,6 +46,39 @@ List<String> scopes = extractScopes(token);
...
 
+### Handling revoked authentication 
+
+When a user's authentication is revoked, the server must be notified to respect the changes (e.g. to close method streams). Invoke the `session.messages.authenticationRevoked` method and raise the appropriate event to notify the server.
+
+```dart
+var userId = 1;
+var removedScopes = ['write'];
+var message = RevokedAuthenticationScope(
+  scopes: removedScopes,
```
hampuslavin, line 57: @SandPod As you said yourself, maybe rename this to `revokedScopes` to be completely consistent with language 🎉

hampuslavin (approved):
Nice! 🍰 One nit that you mentioned in your comment but maybe forgot to update

docs/06-concepts/11-authentication/05-custom-overrides.md
SandPod: Good catch! 🎉

PR #160, opened 2024-09-02
feat: Added 'Advanced authentication schemes' documentation to Authentication / Custom overrides section

christerswahn:
Added 'Advanced authentication schemes' documentation to Authentication / Custom overrides section.

### DIFF
--- a/docs/06-concepts/11-authentication/05-custom-overrides.md
+++ b/docs/06-concepts/11-authentication/05-custom-overrides.md
@@ -188,7 +188,7 @@ In other words the default transport implementation is "invisible" to user code.
 
 ### Implementing your own authentication scheme
 
-If you are implementing your own authentication and are using the `basic` scheme, note that this is supported but will be automatically unwrapped i.e. decoded on the server side before behing handed to your `AuthenticationHandler` implementation. It will in this case receive the decoded auth key value after the `basic` scheme name.
+If you are implementing your own authentication and are using the `basic` scheme, note that this is supported but will be automatically unwrapped i.e. decoded on the server side before being handed to your `AuthenticationHandler` implementation. It will in this case receive the decoded auth key value after the `basic` scheme name.
 
 If you are implementing a different authentication scheme, for example OAuth 2 using bearer tokens, you should override the default method `toHeaderValue` of `AuthenticationKeyManager`. This client-side method converts the authentication key to the format that shall be sent as a transport header to the server.
 
--- a/docs/06-concepts/11-authentication/05-custom-overrides.md
+++ b/docs/06-concepts/11-authentication/05-custom-overrides.md
@@ -8,7 +8,7 @@ When running a custom auth integration it is up to you to build the authenticati
 
 ### Token validation
 
-The token validation is performed by providing a custom `authenticationHandler` callback when initializing Serverpod. The callback should return an `AuthenticationInfo` object if the token is valid, otherwise `null`.
+The token validation is performed by providing a custom `AuthenticationHandler` callback when initializing Serverpod. The callback should return an `AuthenticationInfo` object if the token is valid, otherwise `null`.
 
 ```dart
 // Initialize Serverpod and connect it with your generated code.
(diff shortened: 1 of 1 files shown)


docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -177,3 +177,18 @@ var token = await client.authenticationKeyManager?.get();
...
+
+### Advanced authentication schemes
+
+By default Serverpod will pass the authentication token from client to server in accordance with the HTTP "authorization" header standard with the "basic" scheme name and encoding. This is securely transferred as the connection is TLS encrypted.
+
+The default implementation encodes and wraps the user-provided token in a "basic" scheme which is automatically unwrapped on the server side before being handed to the user-provided authentication handler described above. *In other words the default transport implementation is "invisible" to user code.*
+
+If you are implementing another authentication scheme, for example OAuth 2 using bearer tokens, you should override the default method `toHeaderValue` of `AuthenticationKeyManager`. This client-side method converts the authentication key to the format that shall be sent as a transport header to the server.
+
+You will also need to implement the `AuthenticationHandler` accordingly, in order to process that header value server-side.
```
hampuslavin, line 189: Question: Could it be a good idea to provide an actual short example of the case you mentioned?

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -177,3 +177,18 @@ var token = await client.authenticationKeyManager?.get();
 ```
 
 The above example retrieves the token from the key manager and stores it in the `token` variable.
+
+### Advanced authentication schemes
+
+By default Serverpod will pass the authentication token from client to server in accordance with the HTTP "authorization" header standard with the "basic" scheme name and encoding. This is securely transferred as the connection is TLS encrypted.
+
+The default implementation encodes and wraps the user-provided token in a "basic" scheme which is automatically unwrapped on the server side before being handed to the user-provided authentication handler described above. *In other words the default transport implementation is "invisible" to user code.*
```
hampuslavin, line 185: Question: Nothing wrong with using italics to accentuate parts, but I don't think this pattern is used elsewhere in the docs? I think consistency and coherence is very important in docs so that's why I'm wondering.

Usually I think if some parts needs to be extra highlighted we do an "info" or "attention" markdown block

christerswahn: Well if we don't do it elsewhere then I agree we shouldn't here. But to me it doesn't quite fit to be a distinct 'info' block. I suggest just having it in its own paragraph, that seems appropriate emphasis

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -177,3 +177,18 @@ var token = await client.authenticationKeyManager?.get();
 ```
 
 The above example retrieves the token from the key manager and stores it in the `token` variable.
+
+### Advanced authentication schemes
+
+By default Serverpod will pass the authentication token from client to server in accordance with the HTTP "authorization" header standard with the "basic" scheme name and encoding. This is securely transferred as the connection is TLS encrypted.
```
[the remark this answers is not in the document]
christerswahn: I opted for quotes since these are string literals (literally! :) ) in the transmitted data rather than code. But either way is fine

docs/06-concepts/11-authentication/05-custom-overrides.md
hampuslavin: Agree!

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -177,3 +177,78 @@ var token = await client.authenticationKeyManager?.get();
...
+
+By default Serverpod will pass the authentication token from client to server in accordance with the HTTP `authorization` header standard with the `basic` scheme name and encoding. This is securely transferred as the connection is TLS encrypted.
+
+The default implementation encodes and wraps the user-provided token in a `basic` scheme which is automatically unwrapped on the server side before being handed to the user-provided authentication handler described above.
+
+In other words the default transport implementation is "invisible" to user code.
+
+### Implementing your own authentication scheme
+
+If you are implementing your own authentication and are using the `basic` scheme, note that this is supported but will be automatically unwrapped i.e. decoded on the server side before behing handed to your `AuthenticationHandler` implementation. It will in this case receive the decoded auth key value after the `basic` scheme name.
```
SandPod, line 191: Sugestion: change `...server side before behing handed...` to `...server side before being handed...`

SandPod (approved):
LGTM with one spelling fix! 👍

PR #163, opened 2024-09-10
chore: Bumps version to 2.1.0.

SandPod:
Bumps the documentation version to 2.1.0.

### Comments around second commit.
For some reason docusaurus can't resolve the relative links in old packages anymore. The page roadmap has been removed so it might be something to do with that.

Simples fix was to just reference our new contribution page.


vlidholt (approved):
LGTM 🥳

PR #167, opened 2024-10-02
feat: Add test tools docs

hampuslavin:
## Descrption / reasoning
The idea is to divide into different pages that has different focus areas. Currently I have the following in mind:
- **Get started**: Easily digested get started info, minimum for getting the simplest test to run
- **The basics**: Additional documentation you will absolutely need to know within "the first hour" of writing tests
- **Advanced exmples**: More complex test setup examples. Can still be common cases but not something you necessarily need to write a first test suite.
- **Best practises**: Inspired by Cypress' (e2e web test framework) [best practices page](https://docs.cypress.io/guides/references/best-practices), as inevitably there will always be cases where a user can do a thing in multiple ways. I've found the Cypress page to be tremendously helpful over the years. It also helps to educate and onboard team members who are new to a framework to quickly get into the "philosophy" of the tool.
image

❓ TODO / Question for reviewer:
- Consider what more best practises to add. Maybe some general testing principles that applied outside of Serverpod as well?

### DIFF
--- a/docs/06-concepts/18-testing/01-get-started.md
+++ b/docs/06-concepts/18-testing/01-get-started.md
@@ -135,4 +135,10 @@ docker-compose up --build --detach
 
 By default this starts up both the `development` and `test` profiles. To only start one profile, simply add `--profile test` to the command.
 
-Now the test is ready to be run!
+Now the test is ready to be run:
+
+```bash
+dart test integration_test
+```
+
+Happy testing!
--- a/docs/06-concepts/18-testing/01-get-started.md
+++ b/docs/06-concepts/18-testing/01-get-started.md
@@ -2,10 +2,16 @@
 
 Serverpod provides simple but feature rich test tools to make testing your backend a breeze.
 
+:::info
+
+For Serverpod Mini projects, everything related to the database in this guide can be ignored.
+
+:::
+
 <details>
 <summary> Have an existing project? Follow these steps first!</summary>
 <p>
-For existing projects, a few extra things need to be done:
+For existing non-Mini projects, a few extra things need to be done:
 1. Add the `server_test_tools_path` key to `config/generator.yaml`. Without this key, the test tools file is not generated. The default location for the generated file is `integration_test/test_tools/serverpod_test_tools.dart`, but this can be set to any path (though should be outside of `lib` as per Dart's test conventions).
 
 2. New projects now come with a test profile in `docker-compose.yaml`. This is not strictly mandatory, but is recommended to ensure that the testing state is never polluted. Add the snippet below to the `docker-compose.yaml` file in the server directory:
@@ -157,7 +163,7 @@ That's it, the project setup should be ready to start using the test tools!
 </p>
 </details>
 
-Go to the server directory and generate the test tools by running `serverpod generate`. The default location for the generated file is `integration_test/test_tools/serverpod_test_tools.dart`. The folder name `integration_test` is chosen to differentiate from unit tests (see the [best practises section](best-practises#unit-integration) for more information on this).
+Go to the server directory and generate the test tools by running `serverpod generate --experimental-features testTools`. The default location for the generated file is `integration_test/test_tools/serverpod_test_tools.dart`. The folder name `integration_test` is chosen to differentiate from unit tests (see the [best practises section](best-practises#unit-integration) for more information on this).
 
 The generated file exports a `withServerpod` helper that enables you to call your endpoints directly like regular functions:
 
--- a/docs/06-concepts/18-testing/01-get-started.md
+++ b/docs/06-concepts/18-testing/01-get-started.md
@@ -118,7 +118,7 @@ A few things to note from the above example:
 
 - The test tools should be imported from the generated test tools file and not the `serverpod_test` package.
 - The `withServerpod` callback takes two parameters: `sessionBuilder` and `endpoints`.
-  - `sessionBuilder` is used to build a `session` object that represents the state of the world for your endpoints and is used to set up scenarios.
+  - `sessionBuilder` is used to build a `session` object that represents the state of the world during an endpoint call and is used to set up scenarios.
   - `endpoints` contains all your Serverpod endpoints and lets you call them.
 
 :::info
--- a/docs/06-concepts/18-testing/01-get-started.md
+++ b/docs/06-concepts/18-testing/01-get-started.md
@@ -37,6 +37,68 @@ redis_test:
     - test
 ```
 
+<details>
+<summary>Or copy the complete file here.</summary>
+<p>
+
+```yaml
+services:
+  # Development services
+  postgres:
+    image: postgres:16.3
+    ports:
+      - '8090:5432'
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_DB: projectname
+      POSTGRES_PASSWORD: "<insert database development password>"
+    volumes:
+      - projectname_data:/var/lib/postgresql/data
+    profiles:
+      - '' # Default profile
+      - dev
+  redis:
+    image: redis:6.2.6
+    ports:
+      - '8091:6379'
+    command: redis-server --requirepass "<insert redis development password>"
+    environment:
+      - REDIS_REPLICATION_MODE=master
+    profiles:
+      - '' # Default profile
+      - dev
+
+  # Test services
+  postgres_test:
+    image: postgres:16.3
+    ports:
+      - '9090:5432'
+    environment:
+      POSTGRES_USER: postgres_test
+      POSTGRES_DB: projectname_test
+      POSTGRES_PASSWORD: "<insert database test password>"
+    volumes:
+      - projectname_data:/var/lib/postgresql/data
+    profiles:
+      - '' # Default profile
+      - test
+  redis_test:
+    image: redis:6.2.6
+    ports:
+      - '9091:6379'
+    command: redis-server --requirepass "<insert redis test password>"
+    environment:
+      - REDIS_REPLICATION_MODE=master
+    profiles:
+      - '' # Default profile
+      - test
+
+volumes:
+  projectname_data:
+```
+
+</p>
+</details>
 3. Create a `test.yaml` file and add it to the `config` directory:
 
 ```yaml
--- a/docs/06-concepts/18-testing/01-get-started.md
+++ b/docs/06-concepts/18-testing/01-get-started.md
@@ -163,7 +163,7 @@ That's it, the project setup should be ready to start using the test tools!
 </p>
 </details>
 
-Go to the server directory and generate the test tools by running `serverpod generate --experimental-features testTools`. The default location for the generated file is `integration_test/test_tools/serverpod_test_tools.dart`. The folder name `integration_test` is chosen to differentiate from unit tests (see the [best practises section](best-practises#unit-integration) for more information on this).
+Go to the server directory and generate the test tools by running `serverpod generate --experimental-features testTools`. The default location for the generated file is `integration_test/test_tools/serverpod_test_tools.dart`. The folder name `integration_test` is chosen to differentiate from unit tests (see the [best practises section](best-practises#unit-and-integration-tests) for more information on this).
 
 The generated file exports a `withServerpod` helper that enables you to call your endpoints directly like regular functions:
 
--- a/docs/06-concepts/18-testing/01-get-started.md
+++ b/docs/06-concepts/18-testing/01-get-started.md
@@ -1,0 +1,138 @@
+# Get started
+
+Serverpod provides simple but feature rich test tools to make testing your backend a breeze.
+
+<details>
+<summary> Have an existing project? Follow these steps first!</summary>
+<p>
+For existing projects, a few extra things need to be done:
+1. Add the `server_test_tools_path` key to `config/generator.yaml`. Without this key, the test tools file is not generated. The default location for the generated file is `integration_test/test_tools/serverpod_test_tools.dart`, but this can be set to any path (though should be outside of `lib` as per Dart's test conventions).
+
+2. New projects now come with a test profile in `docker-compose.yaml`. This is not strictly mandatory, but is recommended to ensure that the testing state is never polluted. Add the snippet below to the `docker-compose.yaml` file in the server directory:
+
+```yaml
+# Test services
+postgres_test:
+  image: postgres:16.3
+  ports:
+    - '9090:5432'
+  environment:
+    POSTGRES_USER: postgres_test
+    POSTGRES_DB: projectname_test
+    POSTGRES_PASSWORD: "<insert database test password>"
+  volumes:
+    - projectname_data:/var/lib/postgresql/data
+  profiles:
+    - '' # Default profile
+    - test
+redis_test:
+  image: redis:6.2.6
+  ports:
+    - '9091:6379'
+  command: redis-server --requirepass "<insert redis test password>"
+  environment:
+    - REDIS_REPLICATION_MODE=master
+  profiles:
+    - '' # Default profile
+    - test
+```
+
+3. Create a `test.yaml` file and add it to the `config` directory:
+
+```yaml
+# This is the configuration file for your local test environment. By
+# default, it runs a single server on port 8090. To set up your server, you will
+# need to add the name of the database you are connecting to and the user name.
+# The password for the database is stored in the config/passwords.yaml.
+#
+# When running your server locally, the server ports are the same as the public
+# facing ports.
+
+# Configuration for the main API test server.
+apiServer:
+  port: 9080
+  publicHost: localhost
+  publicPort: 9080
+  publicScheme: http
+
+# Configuration for the Insights test server.
+insightsServer:
+  port: 9081
+  publicHost: localhost
+  publicPort: 9081
+  publicScheme: http
+
+# Configuration for the web test server.
+webServer:
+  port: 9082
+  publicHost: localhost
+  publicPort: 9082
+  publicScheme: http
+
+# This is the database setup for your test server.
+database:
+  host: localhost
+  port: 9090
+  name: projectname_test
+  user: postgres
+
+# This is the setup for your Redis test instance.
+redis:
+  enabled: false
+  host: localhost
+  port: 9091
+```
+
+4. Add this entry to `config/passwords.yaml`
+
+```yaml
+test:
+  database: '<insert database test password>'
+  redis: '<insert redis test password>'
+```
+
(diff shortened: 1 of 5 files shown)


docs/06-concepts/18-testing.md
```diff
@@ -0,0 +1,365 @@
...
+Serverpod provides simple but feature rich test tools to make testing your backend a breeze.
+
+## Getting started
+
+Go to the server directory and generate the test tools by running `serverpod generate`. The default location for the generated file is `integration_test/test_tools/serverpod_test_tools.dart`.
+
+The generated file exports a `withServerpod` helper that enables you to call your endpoints directly like regular functions:
+
+```dart
+// Import the generated file, it contains everything you need.
```
christerswahn, line 12: Since this quote starts with an import it gives the impression it is a complete file. Suggest you make it complete with main() etc as well:

```
import 'package:test/test.dart';
import '../../test_tools/serverpod_test_tools.dart';

void main() {
  withServerpod('Given Example endpoint', (endpoints, session) {
    test('when calling `hello` then should return greeting', () async {
      final greeting = await endpoints.example.hello(session, 'Michael');
      expect(greeting, 'Hello, Michael!');
    });
  });
}

```

hampuslavin: You're completely right, missed that! 👍

hampuslavin: Fixed!

docs/06-concepts/18-testing/01-get-started.md
```diff
@@ -0,0 +1,138 @@
...
+
+Serverpod provides simple but feature rich test tools to make testing your backend a breeze.
+
+<details>
+<summary> Have an existing project? Follow these steps first!</summary>
+<p>
+For existing projects, a few extra things need to be done:
+1. Add the `server_test_tools_path` key to `config/generator.yaml`. Without this key, the test tools file is not generated. The default location for the generated file is `integration_test/test_tools/serverpod_test_tools.dart`, but this can be set to any path (though should be outside of `lib` as per Dart's test conventions).
+
+2. New projects now come with a test profile in `docker-compose.yaml`. This is not strictly mandatory, but is recommended to ensure that the testing state is never polluted. Add the snippet below to the `docker-compose.yaml` file in the server directory:
```
SandPod, line 11: Suggestion: Also link to the complete file after the modifications.

docs/06-concepts/18-testing/01-get-started.md
```diff
@@ -0,0 +1,138 @@
...
+    });
+  });
+}
+```
+
+A few things to note from the above example:
+
+- The test tools should be imported from the generated test tools file and not the `serverpod_test` package.
+- The `withServerpod` callback takes two parameters: `sessionBuilder` and `endpoints`.
+  - `sessionBuilder` is used to build a `session` object that represents the state of the world during an endpoint call and is used to set up scenarios.
```
SandPod, line 121: Suggestion(NIT): Use another word than "world". Maybe "server state", "session state" or "server and session state"?

docs/06-concepts/18-testing/01-get-started.md
```diff
@@ -0,0 +1,138 @@
...
+  - `endpoints` contains all your Serverpod endpoints and lets you call them.
```
SandPod, line 122: Question: Does this also contain your module endpoints?

docs/06-concepts/18-testing/01-get-started.md
```diff
@@ -0,0 +1,138 @@
...
+
+Before you can run the test you also need to start the Postgres and Redis:
+
+```bash
+docker-compose up --build --detach
+```
+
+By default this starts up both the `development` and `test` profiles. To only start one profile, simply add `--profile test` to the command.
+
+Now the test is ready to be run!
```
SandPod, line 138: Suggestion: End the tutorial by running the test also.

Since we are using a different path for our tests we might as well include that in the tutorial as well.

docs/06-concepts/18-testing/01-get-started.md
```diff
@@ -0,0 +1,138 @@
...
+test:
+  database: '<insert database test password>'
+  redis: '<insert redis test password>'
+```
+
+That's it, the project setup should be ready to start using the test tools!
+</p>
+</details>
+
+Go to the server directory and generate the test tools by running `serverpod generate`. The default location for the generated file is `integration_test/test_tools/serverpod_test_tools.dart`.
```
SandPod, line 98: Question: Should we include a small motivation for why we don't generate this inside of the normal "test" folder? The reason for separating the integration tests from the unit tests.

SandPod: Maybe that is more for the best practices section?

docs/06-concepts/18-testing/02-the-basics.md
```diff
@@ -0,0 +1,287 @@
...
+}
+```
+
+To create a new state, simply call `copyWith` with the new properties and use the new session builder in the endpoint call.
+
+### Setting authenticated state
+
+To control the authenticated state of the session, the `AuthenticationOverride` class can be used:
+
+```dart
```
SandPod, line 31: Suggestion/Discussion: Don't expose the implementation details of the TestSessionBuilder in the docs.

The implementation details for the TestSessionBuilder are going to evolve over time. Will we keep up with the documentation? Maybe we can explain it more abstract and then go into single details for each "copyWith" capability instead?

docs/06-concepts/18-testing/02-the-basics.md
```diff
@@ -0,0 +1,287 @@
...
+  /// Sets the session to be unauthenticated. This is the default.
+  static AuthenticationOverride unauthenticated();
+}
+```
+
+Pass it to the `sessionBuilder.copyWith` to simulate different scenarios. Below follows an example for each case:
+
+```dart
+withServerpod('Given AuthenticatedExample endpoint',
+    (sessionBuilder, endpoints) {
```
SandPod, line 48: Suggestion(NIT): Can you force these to be on the same line?

docs/06-concepts/18-testing/02-the-basics.md
```diff
@@ -0,0 +1,287 @@
...
+
+:::
+
+```dart
+withServerpod('Given Products endpoint when authenticated',
+    (sessionBuilder, endpoints) {
+  const int userId = 1234;
+  var authenticatedSession = sessionBuilder
+      .copyWith(
+        authentication: AuthenticationOverride.authenticationInfo(
```
SandPod, line 98: Question: Is the authentication important for this example or could it be removed? Seems like redundant information to prove the seeding point.

docs/06-concepts/18-testing/02-the-basics.md
```diff
@@ -0,0 +1,287 @@
...
+
+The following optional configuration options are available to pass as a second argument to `withServerpod`:
+
+```dart
+{
+  RollbackDatabase? rollbackDatabase = RollbackDatabase.afterEach,
+  String? runMode = ServerpodRunmode.test,
+  bool? enableSessionLogging = false,
+  bool? applyMigrations = true,
+}
```
SandPod, line 146: Suggestion: Should we present these in a table instead?

Feels like that is easier on the eye to parse and provide an overview?

SandPod: We have used tables in other sections of our documentation. Might be worth keeping this consistent.

docs/06-concepts/18-testing/02-the-basics.md
```diff
@@ -0,0 +1,287 @@
...
+
+### `enableSessionLogging`
+
+Wether session logging should be enabled. Defaults to `false`.
+
+### `applyMigrations`
+
+Wether pending migrations should be applied when starting Serverpod. Defaults to `true`.
+
+## Test exceptions
```
SandPod, line 224: Suggestion: Use a table for the exceptions.

Same feedback here, a table gives a clear overview of the options and then we don't need to remember to sync the implementation details with the copied code here.

docs/06-concepts/18-testing/04-best-practises.md
```diff
@@ -0,0 +1,103 @@
...
+
+```dart
+// Only import the generated test tools file.
+// It re-exports all helpers and types that are needed.
+import 'serverpod_test_tools.dart'; ✅ 
+```
+
+### Database clean up
+
+Unless configured otherwise, by default `withServerpod` does all database operations inside a transaction that is rolled back after each `test`.
```
SandPod, line 23: Suggestion: Mention the configuration option that changes this behavior, if someone has not read that section of the documentation it can be good to point them there.

SandPod (commented):
Awesome work!

I like the structure presented here, have a couple of comments for different sections.
Ping me if anything is unclear!

docs/06-concepts/18-testing/01-get-started.md
hampuslavin: Yes, great point!

docs/06-concepts/18-testing/01-get-started.md
hampuslavin: Hmmm, great point. Could be both 🤷 
image

docs/06-concepts/18-testing/02-the-basics.md
hampuslavin: Dart formatter don't want it 😢

docs/06-concepts/18-testing/02-the-basics.md
hampuslavin: Great catch, we should absolutely remove that 👍

docs/06-concepts/18-testing/02-the-basics.md
hampuslavin: If we use tables in other places it makes sense to be consistent 👍

docs/06-concepts/18-testing/01-get-started.md
hampuslavin: Can you clarify what you mean? Adding `dart test ..` or?

docs/06-concepts/18-testing/01-get-started.md
hampuslavin: Nope, I think the philosophy at least in this first iteration should be that the module should have its own tests (using test tools). If this is too limiting we can add it of course. So that said, I'll add a note about that 👍

docs/06-concepts/18-testing/01-get-started.md
hampuslavin: Do you mean linking to the file in the Serverpod repo?

hampuslavin: Then I'll have to wait for the templates PR to be merged first.

docs/06-concepts/18-testing/02-the-basics.md
hampuslavin: I would argue that this is not implementation details as I'm only showing the public interface (without any implementation code). However we phrase this part - if the public interface changes that will inevitable require us to sync with the docs.

That said, maybe I can dissect this an lift out the comments as regular documentation text and then have the functions as snippets. E.g. something along these lines: 
>  `copyWith` creates a new unique session with the provided properties. This is useful for setting up different session states in the tests or simulating multiple users.
```dart
  TestSessionBuilder copyWith({
    AuthenticationOverride? authentication,
    bool? enableLogging,
  });
```

docs/06-concepts/18-testing/01-get-started.md
SandPod: Yes, or possibly have it as a foldout in the foldout? Just so that it is possible to do a complete copy paste of the file if required.

docs/06-concepts/18-testing/01-get-started.md
SandPod: It might make sense to include these in your generated endpoints. Since they essentially are just normal endpoints right?

docs/06-concepts/18-testing/01-get-started.md
SandPod: Yes, exactly. Since the user can't just run `dart test` to run these tests but need to specify the path to the `integrat_test` folder we should have it as a step.

So `dart test integration_test`. and then maybe in the advanced examples include information about `--concurrency=1` when preserving state in the database between tests.

docs/06-concepts/18-testing/04-best-practises.md
```diff
@@ -0,0 +1,116 @@
+# Best practises
+
+## Imports
```
SandPod, line 3: suggestion: Add a description for the rule.

Does it make sense to always have a description for the rule?

Same for the new `Unit and integration tests {#unit-integration}`. Instead of having the general description in the "Don't" place it at the top.

docs/06-concepts/18-testing/02-the-basics.md
SandPod: You are correct, this is the interface, but the comments we have there and the structure of the code is implementation details. So I think it makes more sense to highlight the specific interfaces and keep away from documenting class declarations.

docs/06-concepts/18-testing/02-the-basics.md
```diff
@@ -0,0 +1,242 @@
...
+
+The following exceptions are exported from the generated test tools file and can be thrown in various scenarios, see below.
```
SandPod, line 214: Suggestion: Rephrase to more reflect these are the exceptions that can be thrown by the test framework during a method call.

For example: `The following exceptions are used by the test framework....`

SandPod: Should the transaction exception be part of this list as well?

SandPod (commented):
Some more minor NITs, do you want to convert this to a PR instead of it being a draft?

docs/06-concepts/18-testing/02-the-basics.md
hampuslavin: Since the different exceptions are thrown in different scenarios, I wrote those things on each exception. The transaction case is throwing an `InvalidConfigurationException`. This is a general exception, but it is also documented in the concurrent transacations section that this is what will be thrown.

docs/06-concepts/18-testing/01-get-started.md
hampuslavin: Tracked here, we may or may not wait for its PR before merging in: https://github.com/serverpod/serverpod/issues/2830

docs/06-concepts/18-testing/02-the-basics.md
hampuslavin: Made small clarification: 49270cc

docs/06-concepts/18-testing/01-get-started.md
hampuslavin: Makes sense!

docs/06-concepts/18-testing/02-the-basics.md
hampuslavin: Tried a rework here, but it's a bit tricky. Let me know what you think: 049c08d

docs/06-concepts/18-testing/02-the-basics.md
```diff
@@ -0,0 +1,248 @@
+# The basics
+
+## Using `sessionBuilder` to set up a test scenario
```
SandPod, line 3: Suggestion: Shorten to `Set up test scenario`.

Menu looks a bit cramped with this long of a headline.

docs/06-concepts/18-testing/04-best-practises.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Do
+
+```dart
+// Only import the generated test tools file.
+// It re-exports all helpers and types that are needed.
+import 'serverpod_test_tools.dart'; ✅ 
+```
+
+### Database clean up
```
SandPod, line 23: Suggestion: Change this to only use 2 `#`.

The right side menu on this page is a bit funky, do we really need "do" and "don't" there? maybe we could use 4 `#` to remove them? Or I think there is another way to exclude something from the menu also.

docs/06-concepts/18-testing/02-the-basics.md
```diff
@@ -0,0 +1,248 @@
...
+      final future = endpoints.authenticatedExample
+          .hello(unauthenticatedSessionBuilder, 'Michael');
+      await expectLater(
+          future, throwsA(isA<ServerpodUnauthenticatedException>()));
+    });
+  });
+});
+```
+
+### Seeding the database {#seeding-the-database}
```
SandPod, line 79: Question: Is the additional `{#seeding-the-database}` really required here?

SandPod: As long as there are no collisions I think this name is automatically created.

If it is not required, please check this for all headers that have this addition.

SandPod (commented):
Couple of minor last comments!

Then I think this is good enough to release :)

docs/06-concepts/18-testing/02-the-basics.md
hampuslavin: Oh, I didn't realize it was auto-generated. Will check all of these 👍

docs/06-concepts/18-testing/04-best-practises.md
hampuslavin: I'll check if there's a way to hide them from the menu!

docs/06-concepts/18-testing/03-advanced-examples.md
```diff
@@ -0,0 +1,98 @@
+# Advanced examples
+
+## Test business logic that depends on `Session` {#business-logic-depends-on-session}
+
+It is common to break out business logic into modules and keep it separate from the endpoints. If such a module depends on a `Session` object (e.g to interact with the database), then the `withServerpod` helper can still be used and the second `endpoint` argument can simply be ignored:
+
+```dart
+withServerpod('Given decreasing product quantity when quantity is zero', (
+  sessionBuilder,
+  _ /* Ignore */,
```
vlidholt, line 10: I think the that `_` is generally known to be ignored (I'd propose we skip the comment here).

hampuslavin: Good point!

docs/06-concepts/18-testing/04-best-practises.md
```diff
@@ -1,3 +1,8 @@
+---
+# Don't display do's and don'ts in the table of contents
+toc_max_heading_level: 2
```
SandPod, line 3: Nice! 🙌

SandPod (approved):
Great work!

I think this looks good to me! :D Let's test it out and see if people understand it :)

PR #168, opened 2024-10-04
Update 01-setup.md

phashish:
Fixed the authentication doc from

| **userCanEditFullName** | True if users can view their user name. | true |

to 

| **userCanEditFullName** | True if users can edit their full name. | true |

### DIFF
--- a/versioned_docs/version-2.1.0/06-concepts/11-authentication/01-setup.md
+++ b/versioned_docs/version-2.1.0/06-concepts/11-authentication/01-setup.md
@@ -97,7 +97,7 @@ void run(List<String> args) async {
 | **emailSignInFailureResetTime** | The reset period for email sign in attempts. Defaults to 5 minutes.| 5min |
 | **userCanEditUserImage** | True if users can update their profile images.| true |
 | **userCanEditUserName** | True if users can edit their user names. | true |
-| **userCanEditFullName** | True if users can view their user name. | true | 
+| **userCanEditFullName** | True if users can edit their full name. | true | 
 | **userCanSeeUserName** | True if users can view their user name. | true |
 | **userCanSeeFullName** | True if users can view their full name. | true |
 | **enableUserImages** | True if user images are enabled. | true | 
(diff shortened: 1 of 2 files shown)


versioned_docs/version-2.1.0/06-concepts/11-authentication/01-setup.md
```diff
@@ -97,7 +97,7 @@ void run(List<String> args) async {
 | **emailSignInFailureResetTime** | The reset period for email sign in attempts. Defaults to 5 minutes.| 5min |
 | **userCanEditUserImage** | True if users can update their profile images.| true |
 | **userCanEditUserName** | True if users can edit their user names. | true |
-| **userCanEditFullName** | True if users can view their user name. | true |
```
SandPod, line 100: Suggestion: Apply this change to the non versioned docs as well.

If you search for this line you can modify it in all places.

SandPod (changes requested):
Thank you for fixing this up! 🙏 

Could you please apply this fix to all versions of the documentation?
It would unfortunately be reverted once we release a new version of our documentation the way it is applied now.

versioned_docs/version-2.1.0/06-concepts/11-authentication/01-setup.md
phashish: Sure, on it.

phashish: https://github.com/serverpod/serverpod_docs/pull/169

SandPod: Thanks for being so quick on this. Could you add that change as a commit to this PR instead? Then we have the full fix as one change in main later.

phashish: Done. Please check.

SandPod (approved):
LGTM! 👍

PR #172, opened 2024-10-05
cleanup/clarification of relation-queries

macMikey:
* fix typos in the like example
* make pagination clearer by making the page size correspond to the offset


hampuslavin (approved):
Nice! 🥇

PR #179, opened 2024-10-10
docs: Adding documentation for session logs improvement

klkucaj:


docs/06-concepts/07-configuration.md
```diff
@@ -31,6 +31,8 @@ There are three different ways to configure Serverpod: with environment variable
 | SERVERPOD_REDIS_USER                         | redis.user                           | -         | The user name for Redis authentication           |
 | SERVERPOD_REDIS_ENABLED                      | redis.enabled                        | false     | Indicates if Redis is enabled                    |
 | SERVERPOD_MAX_REQUEST_SIZE                   | maxRequestSize                       | 524288    | The maximum size of requests allowed in bytes    |
+| SERVERPOD_SESSION_PERSISTENT_LOG_ENABLED     | sessionLogs.persistentEnabled        | false     | Enables or disables logging session data to the database. If no database config is provided, it will default to `false`. |
+| SERVERPOD_SESSION_CONSOLE_LOG_ENABLED        | sessionLogs.consoleEnabled           | false     | Enables or disables logging session data to the console. |
```
Isakdl, line 35: Question: Is this true? That we turn off all logging by default? 

The previous behavior was that database logging was enabled always unless we were running Serverpod mini.

klkucaj: @Isakdl yes, if the configs are not present or false, it will not write to db or print to console

Isakdl: Hmm, that would be a breaking change, I think that if no settings are supplied at all then database logging should be enabled and console logging disabled. Just to not change the behavior from now. If we merge this now then we will stop logging in people's production environments unless they update their config files.

klkucaj: @Isakdl okay will update both, this pr and changes

klkucaj: @Isakdl done!

Isakdl (approved):
LGTM

PR #181, opened 2024-10-13
fix: Use new auth object in session.

gazialankus:
Similar fix to issue #115


hampuslavin (dismissed):
@gazialankus Actually, would you mind just updating the paragraph above this change to reflect your change? ☺️

gazialankus:
Ah I completely missed that, thank you! I was about to call it a property but [Dart docs](https://dart.dev/language/methods#getters-and-setters) calls it a getter, I hope this is the right way to call it.

PR #182, opened 2024-10-13
fix: Correct wrong name of authenticationInfo variable.

gazialankus:
Variable names did not match


hampuslavin (approved):
Good catch, thanks! 🙏

PR #188, opened 2024-10-16
fix: Add missing step for existing project

hampuslavin:
## Changes
- Main change is to add the missing step to add `serverpod_test` package
- Changes the example test to actually pass for the default example endpoint
- Small reformats to make it more user friendly

### DIFF
--- a/docs/06-concepts/18-testing/01-get-started.md
+++ b/docs/06-concepts/18-testing/01-get-started.md
@@ -18,22 +18,28 @@ For Serverpod Mini projects, everything related to the database in this guide ca
 <summary> Have an existing project? Follow these steps first!</summary>
 <p>
 For existing non-Mini projects, a few extra things need to be done:
-1. Add the `server_test_tools_path` key to `config/generator.yaml`. Without this key, the test tools file is not generated. The default location for the generated file is `test/integration/test_tools/serverpod_test_tools.dart`, but this can be set to any path (though should be outside of `lib` as per Dart's test conventions).
+1. Add the `server_test_tools_path` key with the value `test/integration/test_tools` to `config/generator.yaml`:
+
+```yaml
+server_test_tools_path: test/integration/test_tools
+```
+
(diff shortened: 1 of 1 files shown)


docs/06-concepts/18-testing/01-get-started.md
```diff
@@ -170,24 +180,39 @@ tags:
 
 ```
 
+6. Finally, add the `test` and `serverpod_test` packages as dev dependencies in `pubspec.yaml`:
+
+```yaml
+dev_dependencies:
+  serverpod_test: 2.1.4 # Should be same version as the `serverpod` package
+  test: '>=1.0.0 <2.0.0'
```
hampuslavin, line 188: Does writing it like this make sense?

SandPod: If this is the Dart test library I would write it with a carrot symbol instead (`^`). A range basically only makes sense if we want to allow multiple major versions.

docs/06-concepts/18-testing/01-get-started.md
```diff
@@ -170,24 +180,39 @@ tags:
 
 ```
 
+6. Finally, add the `test` and `serverpod_test` packages as dev dependencies in `pubspec.yaml`:
+
+```yaml
+dev_dependencies:
+  serverpod_test: 2.1.4 # Should be same version as the `serverpod` package
```
SandPod, line 187: Suggestion: Change this to `<serverpod version>` so that we are compatible with minor versions that this guide is used for.

SandPod (approved):
Two minor things to fix.

docs/06-concepts/18-testing/01-get-started.md
hampuslavin: Good idea!

vlidholt:
I have no additional comments, LGTM 🥳

vlidholt (approved):

PR #189, opened 2024-10-16
fix: Grammar on get started with tests page.

vlidholt:
I'm pretty sure this should be 'is', as 'test tools' is singular (it's just one feature).


hampuslavin (approved):
Whoops, good catch! 👍 🙏

PR #194, opened 2024-10-25
document the sealed feature

BenAuerDev:
This is the documentation part of https://github.com/serverpod/serverpod/pull/2888

To the existing example a h3 and a short intro is added for "Extending a Class", and a h3, intro and example was added explaining the sealed classes. 

Please let me know where there is room for improvement :rocket:

### DIFF
--- a/docs/06-concepts/19-experimental.md
+++ b/docs/06-concepts/19-experimental.md
@@ -24,7 +24,7 @@ The current options you can pass are:
 
 Inheritance allows you to define class hierarchies in your model files by sharing fields between parent and child classes, simplifying class structures and promoting consistency by avoiding duplicate field definitions.
 
-### Extending a  Class
+### Extending a Class
 
 To inherit from a class, use the `extends` keyword in your model files, as shown below:
 
--- a/docs/06-concepts/19-experimental.md
+++ b/docs/06-concepts/19-experimental.md
@@ -24,6 +24,10 @@ The current options you can pass are:
 
 Inheritance allows you to define class hierarchies in your model files by sharing fields between parent and child classes, simplifying class structures and promoting consistency by avoiding duplicate field definitions.
 
+### Extending a  Class
+
+To inherit from a class, use the `extends` keyword in your model files, as shown below:
+
 ```yaml
 class: ParentClass
 fields:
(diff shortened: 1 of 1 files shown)


docs/06-concepts/19-experimental.md
```diff
@@ -46,6 +50,37 @@ class ChildClass extends ParentClass {
 }
 ```
 
+### Sealed Classes
+
+In addition to the `extends` keyword, you can also use the `sealed` keyword to create sealed class hierarchies. A sealed class restricts which other classes can extend or implement it, allowing for more controlled inheritance. For example:
```
SandPod, line 55: Suggestion: Describe the sealed classes allows exchaustive type checking.

> A sealed class restricts which other classes can extend or implement it, allowing for more controlled inheritance. For example

This description doesn't really apply to us in our model case. Better to describe that it allows for exhaustive type checking. I think that is the most common use-case for sealed classes in models.

docs/06-concepts/19-experimental.md
```diff
@@ -46,6 +50,37 @@ class ChildClass extends ParentClass {
...
+
+class ChildClass extends ParentClass {
+    String name;
+    int age;
+}
+```
+
 :::info
-The `extends` keyword does not work for models with a `table` field.
+All files in a sealed hierarchy need to be located in the same sub directory.
```
SandPod, line 85:
```suggestion
All files in a sealed hierarchy need to be located in the same directory.
```

docs/06-concepts/19-experimental.md
```diff
@@ -24,6 +24,10 @@ The current options you can pass are:
 
 Inheritance allows you to define class hierarchies in your model files by sharing fields between parent and child classes, simplifying class structures and promoting consistency by avoiding duplicate field definitions.
```
SandPod, line 25: Suggestion: Add a warning that a new subtype of a class might be a breaking change with older clients.

Regardless of how we solve the serialization/deserialization issue, this is going to be a fact. So we might as well add it as a warning to the documentation.

SandPod (changes requested):
Short and sweet but a couple of comments to fix.

SandPod (approved):
LGTM 👍

PR #195, opened 2024-10-28
fix: Remove the unreferenced List of Employee from the get-started doc

dshukertjr:
Because there is an `employees` field with type `List<Employee>` in the company definition, running `serverpod generate`serverpod generate fails with `The field has an invalid datatype "Employee".` error. 

Later in the docs in the [Object database mapping](https://docs.serverpod.dev/get-started#object-database-mapping) section, there is no employees field, so I'm guessing it's okay to remove it.

### DIFF
--- a/docs/01-get-started.md
+++ b/docs/01-get-started.md
@@ -176,7 +176,6 @@ class: Company
 fields:
   name: String
   foundedDate: DateTime?
-  employees: List<Employee>
 ```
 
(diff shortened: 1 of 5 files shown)


SandPod (changes requested):
Good catch! Could you apply this to the documentation for `1.1.1`, `1.2.0`, `2.0.0` and `2.1.0` as well?

These are located in the `versioned_docs/version-x.x.x` where x.x.x is the version we are targetting.

dshukertjr:
Done ✅

SandPod (approved):
Thank you for fixing this 🙏 👍

PR #210, opened 2024-11-05
add missing equals sign to bash command

BenAuerDev:
Found a missing `=` in the bash command


SandPod (approved):
LTGM!

Thanks for fixing this 🙏

PR #217, opened 2024-11-26
docs: add upgrade to 2.2 page

hampuslavin:
- Adds an upgrade page for Serverpod version 2.2

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -2,3 +2,4 @@ MD001: false
 MD013: false
 MD014: false
 MD024: false
+MD033: false # Allow inline HTML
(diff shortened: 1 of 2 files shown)


.markdownlint.yaml
```diff
@@ -1,4 +1,5 @@
 MD001: false
 MD013: false
 MD014: false
-MD024: false
\ No newline at end of file
+MD024: false
+MD033: false # Allow inline HTML
```
hampuslavin, line 5: Inline HTML is used all over the place so it doesn't make sense to have a linter rule against it.

SandPod (approved):
LGTM! 👍

PR #219, opened 2024-12-03
doc: Adding 'copyWith' method for custom classes

klkucaj:


docs/06-concepts/03-serialization.md
```diff
@@ -24,7 +24,19 @@ For most purposes, you will want to use Serverpod's native serialization. Howeve
     }
     ```
 
-3. You must declare your custom serializable objects in the `config/generator.yaml` file in the server project, the path needs to be accessible from both the server package and the client package.
+3. There must be a method called `copyWith()`, which returns a new instance of the object with the specified fields replaced.
```
SandPod, line 27: Suggestion: Add a hint that it is implemented as a deep copy in the framework and that we recommend implementing it as a deep copy for custom classes as well.

SandPod (changes requested):
Nice addition; please apply this fix to older versions of the documentation as well. (Up untill the point when we added copyWith).

PR #225, opened 2024-12-13
fix: Document free endpoint defintion placement.

SandPod:
Closes: [#3052](https://github.com/serverpod/serverpod/issues/3052)

Updates the documentation to reflect that endpoints can now be placed anywhere in the **lib** directory.
I didn't add any information for the tutorials since these are case examples.

Do we think this is enough, or does it need to be clarified more?

**Changes in get started**
Screenshot 2024-12-13 at 13 28 36

**Changes in working with endpoints**

Screenshot 2024-12-13 at 13 29 23

--------------------------------------

Screenshot 2024-12-13 at 13 29 35


hampuslavin (approved):
🏎️ 🏁

PR #228, opened 2025-01-01
Add more info when changing aws region

itsJoKr:
Wasted an hour because I didn't see that there is `aws_region` field. 

It might take sense to you, but to a newcomer, I think it's weird that docs are talking about changing `ami_instance` and never mention the `aws_region` field.


Isakdl (approved):
LGTM, thank you for the PR.

PR #229, opened 2025-01-02
feat: Add testing contribution guideline

Isakdl:
# Changes

Adds contribution guideline for how we write tests at Serverpod including examples and explanation of our best pracitices.


docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+- **Resilient to Refactoring** – Tests should not break due to internal refactoring. As long as the external behavior remains consistent, tests should pass regardless of code structure changes.
+- **Focused on Behavior, Not Implementation** – We prioritize testing how the code behaves rather than how it is implemented. This prevents unnecessary coupling between tests and production code, fostering long-term stability.
+- **Easy to Maintain and Expand** – Tests should be simple to update or extend as the product evolves. Adding new features should not require widespread changes to existing tests.
+- **Effective at Catching Bugs** – The primary goal of testing is to identify and prevent bugs. Our tests are crafted to cover edge cases, ensure proper functionality, and catch potential regressions.
+
+By adhering to the following principles, we ensure that our test suite remains a valuable and reliable asset as our codebase grows.
+
+This document outlines Serverpod's approach to testing  code. It serves as a guide for writing effective, maintainable, and meaningful tests across all our projects.
```
christerswahn, line 15: Suggestion: Move this paragraph to before the `## Overview` heading

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Key Principles
+
+### 0. Test Independence
+
+- **Tests should be completely independent of one another.**
+- The outcome of a test must never depend on any other test running before or after it.
+- The order in which tests are executed **should not matter.**
+- Running a single test in isolation must produce the same result as running it alongside others.
+- **Exception to the rule:** e2e and integration tests. In scenarios where an external state (like a shared database) is involved, tests may require concurrency mode 1 to prevent interference. But each test should start and end in a clean state.
```
christerswahn, line 25: There are two kinds of independence that can be worth while distinguishing:

- technical independence (reentrancy), which may not be possible with external resources such as DB, in which case non-concurrent execution is required.
- semantic independence, which should only be broken when the tests specifically test a sequence of operations (scenario tests / test scripts). This test scripts need to be "atomic" - they can't be subdivided or change internal order of execution, but should still be independent of each other.

Suggestion: Move the last "exception" point to a separate elaboration, tests ought to be written to be independent even if there are technical dependencies (currently... might be improved!)

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+
+### 1. Clear and Descriptive Test Descriptions
+
+- **Test descriptions should be understandable without reading the test code.**
+- If a test fails, the description alone should make it clear what went wrong.
```
christerswahn, line 30: Suggestion: Add behavior: "make it clear what behavior went wrong"

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+
+- In this case, verifying both the start and end positions of the `SourceSpanException` is essential because they collectively describe the error location, and their correctness is interdependent.
+
+\*Note: SourceSpanException is an object that describes a code error in a source file. See: \*[*https://api.flutter.dev/flutter/package-source\_span\_source\_span/SourceSpanException-class.html*](https://api.flutter.dev/flutter/package-source_span_source_span/SourceSpanException-class.html)&#x20;
+
+### 3. Pure Unit Testing
+
+- **Unit tests should avoid mocking and side effects.**
+- Production code should push side effects **up the call stack**, allowing tests to cover pure methods.
+- **Test the logical feature/unit** rather than a single method or class.
```
christerswahn, line 96: Suggestion: Add "Prefer" to the last point:

- Prefer to **test the logical feature/unit** rather than each single method or class.

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+```
+
+- In this case, verifying both the start and end positions of the `SourceSpanException` is essential because they collectively describe the error location, and their correctness is interdependent.
+
+\*Note: SourceSpanException is an object that describes a code error in a source file. See: \*[*https://api.flutter.dev/flutter/package-source\_span\_source\_span/SourceSpanException-class.html*](https://api.flutter.dev/flutter/package-source_span_source_span/SourceSpanException-class.html)&#x20;
+
+### 3. Pure Unit Testing
+
+- **Unit tests should avoid mocking and side effects.**
+- Production code should push side effects **up the call stack**, allowing tests to cover pure methods.
```
christerswahn, line 95: Suggestion: Add definition of "pure" to this section, since this is for a public audience and not everyone knows exactly what it is, e.g.

**Pure functions**

Pure functions are simply put functions that lack side effects, lack external I/O, and always have the same results for the same inputs. See https://en.wikipedia.org/wiki/Pure_function

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+  return processResponse(response);
+}
+```
+
+- Test `processResponse` directly without mocking the `httpClient`.
+
+### 4. Implementation-Agnostic Tests
+
+- **Do not couple tests to implementation details.**
+- Tests should only break if the behavior changes, not when refactoring code.
```
christerswahn, line 113: Suggestion: Since this is sometimes referred to as black box testing, add a point to spell it out:

- This may be referred to as black box testing.

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+- **Benefits:**
+  - Reduces test brittleness when refactoring.
+  - Ensures tests continue to produce valid objects as the code evolves.
+  - Simplifies object creation without requiring deep lifecycle knowledge.
+
+## Final Thoughts
+
+Consistent application of these principles leads to a robust and maintainable codebase, fostering confidence in our product's reliability and scalability.
+
+We acknowledge that there  are exceptions to the rules. But when deviating from the guidelines mentioned above you have to argue for why. You must have a reasonable argument that aligns with the core goals.
```
christerswahn, line 258: Nit: Remove double space between "there  are"

christerswahn (commented):
Well worded and explicit! Wrote a few suggestions.

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+### 2. Focus on Single Responsibility
+
+- Each test should **only test one thing**.
+- **Avoid** bundling multiple independent checks into a single test.
+- It is acceptable to **repeat the same precondition and action** across tests to ensure each aspect is tested individually.
+
+**Example:**
+
+```dart
+// Good - Tests are split
```
SandPod, line 50: Suggestion: Use the same `Don't` and `Do` pattern used when explaining best practises for our testing framework to ensure consistency in the documentation.

docs/10-contribute/02-testing-guideline.md
SandPod: We might want to rephrase this completely, maybe focus more on the non-side effect part. Since Unit tests have different definitions as well. Or we should explain what we view as a Unit test.

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+### 3. Pure Unit Testing
+
+- **Unit tests should avoid mocking and side effects.**
+- Production code should push side effects **up the call stack**, allowing tests to cover pure methods.
+- **Test the logical feature/unit** rather than a single method or class.
+
+**Example:**
+
+```dart
+// Avoid mocking HTTP requests directly, test the logic that processes the response
```
SandPod, line 101: Suggestion: add a `do` and `don't` to this section as well.

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+}
+```
+
+- Test `processResponse` directly without mocking the `httpClient`.
+
+### 4. Implementation-Agnostic Tests
+
+- **Do not couple tests to implementation details.**
+- Tests should only break if the behavior changes, not when refactoring code.
+- **Unit tests must avoid knowledge of internal methods or variables.**
```
SandPod, line 114: Suggestion/Discussion: Should we add a point that if internal details needs to be tested these are probably justified to be broken out into their own library gaining a public interface that is promised?

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+- **Resilient to Refactoring** – Tests should not break due to internal refactoring. As long as the external behavior remains consistent, tests should pass regardless of code structure changes.
+- **Focused on Behavior, Not Implementation** – We prioritize testing how the code behaves rather than how it is implemented. This prevents unnecessary coupling between tests and production code, fostering long-term stability.
+- **Easy to Maintain and Expand** – Tests should be simple to update or extend as the product evolves. Adding new features should not require widespread changes to existing tests.
+- **Effective at Catching Bugs** – The primary goal of testing is to identify and prevent bugs. Our tests are crafted to cover edge cases, ensure proper functionality, and catch potential regressions.
+
+By adhering to the following principles, we ensure that our test suite remains a valuable and reliable asset as our codebase grows.
+
+This document outlines Serverpod's approach to testing  code. It serves as a guide for writing effective, maintainable, and meaningful tests across all our projects.
+
+## Key Principles
```
SandPod, line 17: Suggestion: Use capital letter only on the first word.

This needs to be applied to all headers.

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+  test('Number below threshold', () {
+    performValidationTest(3, 'Number must be 5 or greater.');
+  });
+
+  test('Number above range', () {
+    performValidationTest(11, 'Number must not exceed 10.');
+  });
+```
+
+### 7. Beneficial Abstractions - Test Builders
```
SandPod, line 201: Suggestion/Discussion: Should we mention factories here as well? When the built objects are dependent on each other?

docs/10-contribute/02-testing-guideline.md
SandPod, line 258: Suggestion: Change to:  `We acknowledge that there are exceptions to the rules. However, any deviation from the guidelines outlined above must be supported by a well-reasoned argument that aligns with the core objectives.`

docs/10-contribute/02-testing-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+- The outcome of a test must never depend on any other test running before or after it.
+- The order in which tests are executed **should not matter.**
+- Running a single test in isolation must produce the same result as running it alongside others.
+- **Exception to the rule:** e2e and integration tests. In scenarios where an external state (like a shared database) is involved, tests may require concurrency mode 1 to prevent interference. But each test should start and end in a clean state.
+
+### 1. Clear and Descriptive Test Descriptions
+
+- **Test descriptions should be understandable without reading the test code.**
+- If a test fails, the description alone should make it clear what went wrong.
+- **Format:** Descriptions follow the "Given, When, Then" style.
```
SandPod, line 31: Suggestion/Discussion: Should we outline what the different words mean?

docs/10-contribute/02-testing-guideline.md
SandPod: Maybe it would be good to break them apart even? One that outlines the test abstraction level and one that focuses on side effects? This might already be covered in point 4. Implementation-agnostic tests.

SandPod (commented):
Nice writeup! I agree with all of it and only have some minor comments and suggestions around structure and phrasing.

I'm not sure we should actually add this to our documentation thoe. Maybe it would make more sense to place this in the actual project and keep the simple Roadmap and Constributions here?

Then we could also have information on how to run tests there as well?
I'm not sure it makes sense to have information specific to "contributors" outside of the project? What do you think?

docs/10-contribute/02-testing-guideline.md
Isakdl: I don't think this is a prefer thing, but a hard rule. We could define "logical feature/unit" a little more in detail however. A unit could be a single method, but most often it is not.

docs/10-contribute/01-contribute.md
```diff
@@ -5,6 +5,7 @@ Serverpod is built by the community for the community. Pull requests are very mu
 <div style={{ position : 'relative', paddingBottom : '56.25%', height : '0' }}><iframe style={{ position : 'absolute', top : '0', left : '0', width : '100%', height : '100%' }} width="560" height="315" src="https://www.youtube-nocookie.com/embed/V3CqPx4jykE" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe></div>
 
 ## Roadmap
```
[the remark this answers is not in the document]
Isakdl: As discussed offline we will place this in the main repo instead.

docs/10-contribute/02-testing-guideline.md
Isakdl: Skipping this one.

PR #231, opened 2025-01-13
fix: Remove info box as limitation no longer applies

BenAuerDev:
Within: Experimental Features/Inheritance

This PR removes an info box informing about the limitation: _"All files in a sealed hierarchy need to be located in the same directory."_

As of merging: [serverpod/pull/2971](https://github.com/serverpod/serverpod/pull/2971) this is no longer the case :muscle: :rocket:


SandPod (commented):
Awesome, thanks for cleaning this up!

I'll merge this once we do the next release.

If the next release is simply a patch we should add this to the versioned documentation as well.

SandPod (approved):
LGTM! 👍

PR #237, opened 2025-02-20
docs: Experimental exception monitoring feature

christerswahn:
Documentation for the experimental exception monitoring feature introduced in PRs:

https://github.com/serverpod/serverpod/pull/3187
https://github.com/serverpod/serverpod/pull/3249

### DIFF
--- a/docs/03-capabilities.md
+++ b/docs/03-capabilities.md
@@ -44,10 +44,6 @@ Serverpod's future calls replace complicated cron jobs. Call a method anytime in
 
 Monitor the database and external services that you are using. Write custom health checks and get notified when something goes wrong.
 
-## Monitoring and error diagnostics integration
-
-Integrate with external services for monitoring and diagnostics, such as [Sentry](https://sentry.io/), [Highlight](https://www.highlight.io/), and [Datadog](https://www.datadoghq.com/), to get detailed and realtime monitoring and metrics.
-
 ## Easy deployment
 
 Serverpod comes with Terraform scripts for Google Cloud Platform and AWS, making deploying your server very quick. We are still working on scripts for other platforms. Please get in touch with us if you want to [contribute](/contribute).
--- a/docs/03-capabilities.md
+++ b/docs/03-capabilities.md
@@ -44,6 +44,10 @@ Serverpod's future calls replace complicated cron jobs. Call a method anytime in
 
 Monitor the database and external services that you are using. Write custom health checks and get notified when something goes wrong.
 
+## Monitoring and error diagnostics integration
+
+Integrate with external services for monitoring and diagnostics, such as [Sentry](https://sentry.io/), [Highlight](https://www.highlight.io/), and [Datadog](https://www.datadoghq.com/), to get detailed and realtime monitoring and metrics.
+
 ## Easy deployment
 
 Serverpod comes with Terraform scripts for Google Cloud Platform and AWS, making deploying your server very quick. We are still working on scripts for other platforms. Please get in touch with us if you want to [contribute](/contribute).
--- a/docs/06-concepts/20-experimental.md
+++ b/docs/06-concepts/20-experimental.md
@@ -1,12 +1,24 @@
 # Experimental features
 
 :::warning
-Experimental features should not be used in production environments, as their stability is uncertain and they may receive breaking changes in upcoming releases.
+Be cautious when using experimental features in production environments, as their stability is uncertain and they may receive breaking changes in upcoming releases.
 :::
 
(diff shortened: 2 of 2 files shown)


docs/03-capabilities.md
```diff
@@ -44,6 +44,10 @@ Serverpod's future calls replace complicated cron jobs. Call a method anytime in
 
 Monitor the database and external services that you are using. Write custom health checks and get notified when something goes wrong.
 
+## Monitoring and error diagnostics integration
+
+Integrate with external services for monitoring and diagnostics, such as [Sentry](https://sentry.io/), [Highlight](https://www.highlight.io/), and [Datadog](https://www.datadoghq.com/), to get detailed and realtime monitoring and metrics.
+
```
SandPod, line 50: Suggestion(Discussion): Remove this from the highlighted capabilities until we bring the monitoring out of experimental.

We do list task scheduling and the web server here as well. But I think we should probably only list released stable features right?

docs/06-concepts/20-experimental.md
```diff
@@ -87,3 +87,131 @@ class ChildClass extends ParentClass {
...
+
+## Exception monitoring
+
+Serverpod allows you to monitor exceptions in a central and flexible way by using the new diagnostic event handlers.
+These work both for exceptions thrown in application code and from the framework (e.g. server startup or shutdown errors).
+
+This can be used to get all exceptions reported in realtime to services for monitoring and diagnostics,
+such as [Sentry](https://sentry.io/), [Highlight](https://www.highlight.io/), and [Datadog](https://www.datadoghq.com/).
+
+It is easy to implement handlers and to implement custom filters in them.
```
SandPod, line 99:
```suggestion
It is easy to implement handlers and define custom filters within them.
```

Removes the `to` repetition.

docs/06-concepts/20-experimental.md
```diff
@@ -87,3 +87,131 @@ class ChildClass extends ParentClass {
...
+Any number of handlers can be added.
+They are run asynchronously and should not affect the behavior or response times of the server.
+
+These event handlers are for diagnostics only,
+they do not allow any behavior-changing action such as suppressing exceptions or converting them to another type.
+
+### Experimental
```
SandPod, line 106: Suggestion: Remove this from the specific feature and any required additions to the top of the page.

Since this is an "Experimental" page, this requirement should be for all features. Opting into the experiment should be the same as going through the setup. 

The `unstable` part should maybe be added to the top of the page if that is something we would like to use as a pattern. It might be good to clarify the difference since `unstable` is usually implied by `experimental`.

docs/06-concepts/20-experimental.md
```diff
@@ -87,3 +87,131 @@ class ChildClass extends ParentClass {
...
+This is introduced as an experimental feature. The API should currently be regarded as unstable which means there may be breaking changes to it between minor versions of the Serverpod package. To make this clear, the important parameters and methods in the Serverpod class have their names prefixed with `unstable`.
+
+It is opt-in by providing event handlers to the Serverpod constructor,
+and by invoking the new method `unstableSubmitDiagnosticEvent`.
+
+When this feature and its interface has matured, proper names without `unstable` will be introduced. If possible, the `unstable` names will remain for some time as `@deprecated`, and then removed.
+
+### Setup
+
+The Serverpod constructor now accepts `unstableDiagnosticEventHandlers` among its `experimentalFeatures` specification,
```
SandPod, line 117: Suggestion: Add that you enable this feature by adding a  diagnostic event handler.

SandPod (commented):
Couple of suggestions.

Don't know if we should list experimental features as capabilities, even if we historically have done that (future calls and web server..)

docs/03-capabilities.md
christerswahn: This is perhaps what someone looking into Serverpod as an alternative reads. Good to point out things that are coming soon as well!

We could maybe add a section at the bottom with features in beta / experimental phase / coming soon?

SandPod: Yes, visibility around these things is always good.

I have two concerns:
- **Clutter on this page**: Maybe this should be as clean as possible with clarity around what is released; since we don't expect this to be a long-running experiment, we might not lose that much by adding it once it is out. We can market experiments through other channels, but this should maybe mostly reflect the stable features.
- **Right to experiment**: We want to preserve the right to completely cancel any experimental feature by listing it as a capability of the framework. It feels like broadcasting it like this might lock it in place and create a bad sentiment if we abort the experiment. Maybe we can promote the feature to this page (under some experiment section) once we are ready for wide adoption and we know we won't back down?

The second point has bitten me in the ass before. Pragmatically, I don't think we have enough of an audience for this to be an issue at the moment. But might as well determine our approach now.

SandPod (approved):
LGTM! 👍

PR #244, opened 2025-02-28
fix: Add documentation for SERVERPOD_SERVER_ID

SandPod:
Adds documentation for the new SERVERPOD_SERVER_ID environment variable introduced in this PR: https://github.com/serverpod/serverpod/pull/3238

We should make the other command line options equally configurable through environment variables.


christerswahn (approved):
Great!

PR #245, opened 2025-02-28
fix: Add documentation for the new types supported.

SandPod:
Adds documentation for the new data types that are now supported in endpoints and models.

We don't seem to explicitly document what is supported in streams so there is no modification of that here now.

We should add something about that later.


christerswahn (approved):
LGTM

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -55,9 +55,9 @@ You can pass the `--watch` flag to `serverpod generate` to watch for changed fil
...
+There are some limitations to how endpoint methods can be implemented. Parameters and return types can be of type `bool`, `int`, `double`, `String`, `UuidValue`, `Duration`, `DateTime`, `ByteData`, `Uri`, `BigInt`, or generated serializable objects (see next section). A typed `Future` should always be returned. Null safety is supported. When passing a `DateTime` it is always converted to UTC.
 
-You can also pass `List` and `Map` as parameters, but they need to be strictly typed with one of the types mentioned above. For `Map`, the keys must be non-nullable strings. E.g., `Map<String, int?>` is valid, but `Map<int, String>` is not.
+You can also pass `List`, `Map`, `Record` and `Set` as parameters, but they need to be strictly typed with one of the types mentioned above. For `Map`, the keys must be non-nullable strings. E.g., `Map<String, int?>` is valid, but `Map<int, String>` is not.
```
tp, line 60:
```suggestion
You can also pass `List`, `Map`, `Record` and `Set` as parameters, but they need to be strictly typed with one of the types mentioned above.
```

This must’ve been a legacy restriction, but now `Map`s support all kind of types (they just have a different underlying layout during transport).

E.g. this uses various key types https://github.com/serverpod/serverΩpod/blob/main/tests/serverpod_test_server/lib/src/endpoints/map_parameters.dart and we even support `Record` keys now https://github.com/serverpod/serverpod/blob/main/tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart#L174

SandPod: Ooh, good catch! I was so focused on adding the new types I missed this. Thank you 🙏

PR #250, opened 2025-03-12
fix: Update 04-deploying-to-aws.md

damolator:
Should this:
WantedBy=muti-user.target 

Be multi with an 'l'?
WantedBy=multi-user.target

### DIFF
--- a/versioned_docs/version-2.4.0/07-deployments/04-deploying-to-aws.md
+++ b/versioned_docs/version-2.4.0/07-deployments/04-deploying-to-aws.md
@@ -146,7 +146,7 @@ ExecStart=/home/ec2-user/serverpod/active/mypod_server/deploy/aws/scripts/run_se
 Restart=always
 
 [Install]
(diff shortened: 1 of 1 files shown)


SandPod:
@damolator - Yes, you are correct! Could you fix this typo in all released versions of the documentation? :)

damolator (commented):
updated PR title

SandPod:
@damolator - In order to preserve this fix, it also needs to be applied to the `next` version of the documentation. Otherwise, it will be reverted once we create a new version of the documentation. This is located in the ´docs/*` folder.

Please apply it there and in any other version in `versioned_docs/*` where this might apply.

damolator:
I don't know how to apply it to other versions, or any other version in versioned_docs/*. I was trying to provide a quick fix to the docs, but unfortunately don't have time to read through github and serverpod commit docs, and keep reapplying updates.  I'm sure someone on the serverpod team can do this in 5 minutes.

PR #256, opened 2025-03-31
feat: Add tracking

dkbast:
Adds the google analytics property for better insights into which docs pages are most frequently used


SandPod (approved):
LGTM! 🔎

PR #258, opened 2025-04-04
feat: Add documentation for enum defaults.

SandPod:
Adds documentation for enum default values.

Screenshot 2025-04-04 at 12 19 00

### DIFF
--- a/docs/06-concepts/02-models.md
+++ b/docs/06-concepts/02-models.md
@@ -92,6 +92,29 @@ It's recommended to always set `serialized` to `byName` in any new Enum models, 
 
 :::
 
+### Default value
+
+A default value is used when an unknown value is deserialized. This can happen, for example, if a new enum option is added and older clients receive it from the server, or if an enum option is removed but the database still contains the old value.
+
+To configure a default value, use the `default` keyword.
+
+```yaml
+enum: Animal
+serialized: byName
+default: unknown
+values:
+ - unknown
+ - dog
+ - cat
+ - bird
+```
+
+In the example above, if the Enum `Animal` receives an unknown option such as `"fish"` it will be deserialized to `Animal.unknown`. This is useful for maintaining backward compatibility when changing the enum values.
+
+:::warning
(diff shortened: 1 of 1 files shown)


dkbast (approved):

docs/06-concepts/02-models.md
```diff
@@ -92,6 +92,29 @@ It's recommended to always set `serialized` to `byName` in any new Enum models,
...
+ - unknown
+ - dog
+ - cat
+ - bird
+```
+
+In the example above, if the Enum `Animal` receives an unknown option such as `"fish"` it will be deserialized to `Animal.unknown`. This is useful for maintaining backward compatibility when changing the enum values.
+
+:::warning
+If a default value is not specified, deserialization of unknown values will throw an exception. Adding a default value prevents these exceptions, but may also hide real issues in your data. Use this feature with caution.
```
tp, line 115:
```suggestion
If no default value is specified, deserialization of unknown values will throw an exception. Adding a default value prevents these exceptions, but may also hide real issues in your data. Use this feature with caution.
```

Would be my preference, but either way, this is a very important and useful warning :)

(It’s clear to me, just hoping no one gets confused by “unknown” as the concept (i.e. not included in the possible values) vs. the enum case named unknown in the example above. But of the top of my head I also can not think of better names right now, other than maybe “fallback”…

tp (approved):
Nice :) I think it clearly explains the concept and give actionable advice and when (not) to use it.

docs/06-concepts/02-models.md
SandPod: Good catch, could lead to some confusion there. Would `invalid` be more clear?

dkbast: I've seen many projects using "unknown" for this usecase - "invalid" should probably throw instead?

SandPod: Yeah maybe. I'll just keep it as is then and well see if it is unclear :)

PR #259, opened 2025-04-04
fix: Fixes a typo that was fixed in the repo in 2.3.0

SandPod:
This typo was fixed in this PR https://github.com/serverpod/serverpod/pull/2976 and shipped with 2.3.0.

This fixes the typo for the newer versions.

Noticed and originally fixed by @damolator in https://github.com/serverpod/serverpod_docs/pull/250


christerswahn (approved):
LGTM

PR #262, opened 2025-04-07
feat: Adds link to changelog and Makefile

dkbast:
Adds a link to the changelog
Adds a Makefile with the common commands

image

PR #263, opened 2025-04-07
Split getting started

dkbast:
This is plitting the getting started guide to see who actually goes through the docs and is not just scrolling over it, this is the first part in our initiative to rework the documentation

Merge after landing #262


docs/01-get-started/01-overview.md
dkbast, line 1: This file could also be named 01-get-started.md, then it would not have a separate entry in the menu but would be the page thats shown when you click on the "folder" name

docusaurus.config.js
```diff
@@ -138,6 +138,16 @@ const config = {
             // Moved in version 2.1.0
             from: ['/roadmap'],
             to: '/contribute',
+          },            
+          {
+            // Moved in version 2.5.0+2
+            from: ['/get-started'],
+            to: '/get-started/overview',
+          },
```
SandPod, line 146: Comment: This might be redundant but should do no harm.

SandPod (approved):
LGTM! 👍

PR #268, opened 2025-04-15
docs: Add documentation for `Endpoint` inheritance

tp:
* `@ignoreEndpoint` annotation on classes and methods.
* Endpoint inheritance.
* Endpoint inheritance from abstract classes.
* Versioning your API (Inheritance + deprecated annotation + changes in new API).

### DIFF
--- a/docs/06-concepts/01-working-with-endpoints.md
+++ b/docs/06-concepts/01-working-with-endpoints.md
@@ -199,7 +199,7 @@ Since `CalculatorEndpoint` is marked as `@ignoreEndpoint` it will not be exposed
 
 ### Overriding endpoint methods
 
-Suppose you had a `GreeterBaseEndpoint` whose behavior you want to adapt.
+It is possible to override methods of the super class. This can be useful when you want to modify the behavior of specific methods but preserve the rest.
 
 ```dart
 import 'package:serverpod/serverpod.dart';
--- a/docs/06-concepts/01-working-with-endpoints.md
+++ b/docs/06-concepts/01-working-with-endpoints.md
@@ -90,7 +90,7 @@ class ExampleEndpoint extends Endpoint {
 
 The above code will not generate any server or client bindings for the example endpoint.
 
-Alternatively you can disable single methods like in the example below, where `hello` is exposed to the client, but `goodbye` is not:
+Alternatively, you can disable single methods like in the example below, where `hello` is exposed to the client, but `goodbye` is not:
 
 ```dart
 import 'package:serverpod/serverpod.dart';
@@ -109,7 +109,7 @@ class ExampleEndpoint extends Endpoint {
 
 ## Endpoint inheritance
 
-It is possible to extend existing endpoints, from your own app or other modules, by subclassing them. In case the parent endpoint was marked as `abstract` or `@ignoreEndpoint`, no client code is generated for it, but a client will be generated for your subclass – as long as it does not opt-out by any of the aforementioned ways.
+It is possible to extend existing endpoints, from your own app or other modules, by subclassing them. If the parent endpoint was marked as `abstract` or `@ignoreEndpoint`, no client code is generated for it, but a client will be generated for your subclass – as long as it does not opt-out by any of the aforementioned ways.
 
 ```dart
 import 'package:serverpod/serverpod.dart';
@@ -153,14 +153,14 @@ class ExcitedGreeter extends GreeterBase {
 }
 ```
 
-In the above example `ExcitedGreeter` inherits from `GreeterBase`. Since the base class is marked as `abstract` no client was generated for it, but now all of its visible endpoint methods would be exposed through the sub-class `ExcitedGreeter`.
+In the above example, `ExcitedGreeter` inherits from `GreeterBase`. Since the base class is marked as `abstract`, no client was generated for it, but now all of its visible endpoint methods would be exposed through the sub-class `ExcitedGreeter`.
 
 The sub-class `ExcitedGreeter` is an example of all possible modifications to the abstract base class and contains the following changes:
 
 - `hello` is overriden and augments the super class' implementation with a trailing `!!!`
 - `goodbye` is now exposed through the sub-class, as it was not individually hidden and the sub-class does not further augment it
-- `wave`, which is explictly ignored in the base-class is still not exposed, as the sub-class did not opt into re-exposing it
-- `wavePerson` was explicitly overriden to be ignored. Don't worry about the `throw` in the method implementation, which is only needed to satisfy the compiler. In practice this method can not be called from the client anymore.
+- `wave`, which is explicitly ignored in the base-class, is still not exposed, as the sub-class did not opt into re-exposing it
+- `wavePerson` was explicitly overriden to be ignored. Don't worry about the `throw` in the method implementation, which is only needed to satisfy the compiler. In practice, this method can not be called from the client anymore.
 
 ### API versioning for breaking changes
 
@@ -189,12 +189,12 @@ class TeamV2Endpoint extends TeamEndpoint {
 }
 ```
 
-In the above example we created a new `TeamV2` endpoint which hides the `join` method and instead exposes a `joinWithCode` method, plus all the other inherited (and untouched) methods from the parent class.
+In the above example, we created a new `TeamV2` endpoint, which hides the `join` method and instead exposes a `joinWithCode` method, plus all the other inherited (and untouched) methods from the parent class.
 
 While we may have liked to re-use the `join` method name, Dart inheritance rules do not allow doing so.
 
-Then in your client you could move all usages from `client.team` to `client.teamV2`, and could eventually remove the old endpoint. Either by marking it with `@ignoreEndpoint` on the class or deleting it and moving the untouched method implementations you want to keep to the V2 endpoint class.
+In your client, you could then move all usages from `client.team` to `client.teamV2` and eventually remove the old endpoint. Either by marking it with `@ignoreEndpoint` on the class or deleting it and moving the untouched method implementations you want to keep to the V2 endpoint class.
 
 An alternative pattern to consider would be to move all the business logic for an endpoint into a helper class, and then call into that from the endpoint. In case you want to create a V2 version later, you might be able to reuse most of the underlying business logic through that helper class, and don't have to sub-class the old endpoint. This has the added benefit of the endpoint class clearly listing all exposed methods, and you don't have to wonder what you inherit from the base class.
 
-Either approach has its pros and cons, and it depends on the concrete circumstances to pick the one that is most useful. Both give you all the tools you need to extend and update your API while gracefully moving clients along, giving them time to update.
+Either approach has its pros and cons, and it depends on the concrete circumstances to pick the one that is most useful. Both give you all the tools you need to extend and update your API while moving clients along gracefully, giving them time to update.
--- a/docs/06-concepts/01-working-with-endpoints.md
+++ b/docs/06-concepts/01-working-with-endpoints.md
@@ -119,7 +119,7 @@ Endpoints can be based on other endpoints using inheritance, like `class ChildEn
 
 Currently, there are the following possibilities to extend another `Endpoint` class:
 
-### Inheriting from a concrete (visible) `Endpoint` class
+### Inheriting from an `Endpoint` class
 
 Suppose there is an existing `CalculatorEndpoint` (presumably defined in another package, otherwise the existing one could be adapted) which you want to extend with further functionality.
 
@@ -142,7 +142,7 @@ class MyCalculatorEndpoint extends CalculatorEndpoint {
 The generated client code will now be able to access both `CalculatorEndpoint` and `MyCalculatorEndpoint`.
 Whereas the `CalculatorEndpoint` only exposes the original `add` method, `MyCalculatorEndpoint` now exposes both the inherited `add` and its own `subtract` methods.
 
-### Inheriting from an `abstract` `Endpoint` class
+### Inheriting from an `Endpoint` class marked `abstract`
 
 Endpoints marked as `abstract` are not added to the server. But if they are subclassed, their methods will be exposed through the subclass.
 
@@ -177,7 +177,7 @@ class MyCalculatorEndpoint extends CalculatorEndpoint {
 
 In this case it will expose both an `add` and a `subtract` method.
 
-### Inheriting from an `@ignoreEndpoint` `Endpoint` class
+### Inheriting from an `Endpoint` class annotated with `@ignoreEndpoint`
 
 Suppose you had an `Endpoint` class marked with `@ignoreEndpoint` and a subclass that extends it:
 
@@ -251,7 +251,7 @@ Don't worry about the exception in the `subtract` implementation. That is only a
 
 Hiding endpoints from a super class is only appropriate in case the parent `class` is `abstract` or annotated with `@ignoreEndpoint`. Otherwise the method that should be hidden on the child would still be accessible via the parent class.
 
-### Unhiding endpoint methods with `@ignoreEndpoint`
+### Unhiding endpoint methods annotated with `@ignoreEndpoint` in the super class
 
 The reverse of the previous example would be a base endpoint that has a method marked with `@ignoreEndpoint`, which you now want to expose on the subclass.
 
--- a/docs/06-concepts/01-working-with-endpoints.md
+++ b/docs/06-concepts/01-working-with-endpoints.md
@@ -155,7 +155,7 @@ class ExcitedGreeter extends GreeterBase {
 
 In the above example `ExcitedGreeter` inherits from `GreeterBase`. Since the base class is marked as `abstract` no client was generated for it, but now all of its visible endpoint methods would be exposed through the sub-class `ExcitedGreeter`.
 
-The sub-class applies the following modifications though:
+The sub-class `ExcitedGreeter` is an example of all possible modifications to the abstract base class and contains the following changes:
 
 - `hello` is overriden and augments the super class' implementation with a trailing `!!!`
 - `goodbye` is now exposed through the sub-class, as it was not individually hidden and the sub-class does not further augment it
--- a/docs/06-concepts/01-working-with-endpoints.md
+++ b/docs/06-concepts/01-working-with-endpoints.md
@@ -164,7 +164,7 @@ The sub-class applies the following modifications though:
 
 ### API versioning for breaking changes
 
-Endpoint sub-class can be useful when having to do a breaking change on an endpoint, but you want to keep sharing most of it's implementation with the old endpoint.
+An endpoint sub-class can be useful when having to do a breaking change on an endpoint, but you need to keep supporting existing clients. Doing so also allows you to keep sharing most of its implementation with the old endpoint.
 
 Imagine you had a "team" management endpoint where before a user could join if they had an e-mail address ending in the expected domain, but now it should be opened up for anyone to join if they can provide an "invite code".
 
--- a/docs/06-concepts/01-working-with-endpoints.md
+++ b/docs/06-concepts/01-working-with-endpoints.md
@@ -115,13 +115,14 @@ In this case the `ExampleEndpoint` will only expose the `hello` method, whereas 
 
 ## Endpoint method inheritance
 
-Endpoints can be based on other endpoints using inheritance, like `class ChildEndpoint extends ParentEndpoint`. If the parent endpoint was marked as `abstract` or `@ignoreEndpoint`, no client code is generated for it, but a client will be generated for your subclass – as long as it does not opt out again.
+Endpoints can be based on other endpoints using inheritance, like `class ChildEndpoint extends ParentEndpoint`. If the parent endpoint was marked as `abstract` or `@ignoreEndpoint`, no client code is generated for it, but a client will be generated for your subclass – as long as it does not opt out again.  
+Inheritance gives you the possibility to modify the behavior of `Endpoint` classes defined in other Serverpod modules.
 
 Currently, there are the following possibilities to extend another `Endpoint` class:
 
 ### Inheriting from an `Endpoint` class
 
-Suppose there is an existing `CalculatorEndpoint` (presumably defined in another package, otherwise the existing one could be adapted) which you want to extend with further functionality.
+Given an existing `Endpoint` class, it is possible to extend or modify its behavior while retaining the already exposed methods.
 
 ```dart
 import 'package:serverpod/serverpod.dart';
--- a/docs/06-concepts/01-working-with-endpoints.md
+++ b/docs/06-concepts/01-working-with-endpoints.md
@@ -109,7 +109,7 @@ class ExampleEndpoint extends Endpoint {
 
 ## Endpoint inheritance
 
-It is possible to extend existing endpoints, from your own app or other modules, by subclassing them. If the parent endpoint was marked as `abstract` or `@ignoreEndpoint`, no client code is generated for it, but a client will be generated for your subclass – as long as it does not opt-out by any of the aforementioned ways.
+Extending existing endpoints from your own app or other modules is possible by subclassing them. If the parent endpoint was marked as `abstract` or `@ignoreEndpoint`, no client code is generated for it, but a client will be generated for your subclass – as long as it does not opt out by any of the aforementioned ways.
 
 ```dart
 import 'package:serverpod/serverpod.dart';
@@ -139,7 +139,7 @@ class ExcitedGreeter extends GreeterBase {
     return '${super.hello(session, name)}!!!';
   }
 
-  // to enable `wave`, which was marked with `@ignoreEndpoint` on the base class, you can override it on the sub-class and call the `super` implementation. This will create the method on the endpoint.
+  // To enable `wave`, which was marked with `@ignoreEndpoint` on the base class, you can override it on the sub-class and call the `super` implementation. This will create the method on the endpoint.
   // @override
   // Future<String> wave(Session session) async {
   //   return super.wave(session);
@@ -157,8 +157,8 @@ In the above example, `ExcitedGreeter` inherits from `GreeterBase`. Since the ba
 
 The sub-class `ExcitedGreeter` is an example of all possible modifications to the abstract base class and contains the following changes:
 
-- `hello` is overriden and augments the super class' implementation with a trailing `!!!`
-- `goodbye` is now exposed through the sub-class, as it was not individually hidden and the sub-class does not further augment it
+- `hello` is overridden and augments the super class' implementation with a trailing `!!!`
+- `goodbye` is now exposed through the sub-class, as it was not individually hidden, and the sub-class does not further augment it
 - `wave`, which is explicitly ignored in the base-class, is still not exposed, as the sub-class did not opt into re-exposing it
 - `wavePerson` was explicitly overriden to be ignored. Don't worry about the `throw` in the method implementation, which is only needed to satisfy the compiler. In practice, this method can not be called from the client anymore.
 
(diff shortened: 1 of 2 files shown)


tp:
please take a look @SandPod @dkbast 🙏

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -75,11 +75,10 @@ The return type must be a typed Future. Supported return types are the same as f
 
 ## Ignore endpoint definition
 
-If you want the code generator to ignore an endpoint definition, you can annotate the class with `@ignoreEndpoint`, imported from `serverpod_shared/annotations.dart`.  This can be useful if you want to keep the definition in your codebase without generating server or client bindings for it.
```
tp, line 78: Removed the extra import here and in the examples for now, as my current feeling is that we'll re-export the annotations (due to the alternatives not being clear wins for now, maybe something to re-think in v3 if package dependencies should change).

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+    return 'Bye $name';
+  }
+}
+```
+
+## Endpoint inheritance
+
+It is possible to extend existing endpoints, from your own app or other modules, by subclassing them. In case the parent endpoint was marked as `abstract` or `@ignoreEndpoint`, no client code is generated for it, but a client will be generated for your subclass – as long as it does not opt-out by any of the aforementioned ways.
+
+```dart
```
tp, line 114: Do you think the code example are fine to follow to see the changes?

I grouped the many variants in one, as otherwise there's too much boilerplate if hide, show, override / extend all get their own section (and there are still further variants possible and not mentioned here, like extending in a backwards compatible way by adding an optional parameter on the base class etc.).

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+import 'package:serverpod/serverpod.dart';
+
+abstract class GreeterBase extends Endpoint {
```
tp, line 117: Alternatively we could also use some recipe related examples, if that is the new direction 👨‍🍳

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+}
+```
+
+In the above example we created a new `TeamV2` endpoint which hides the `join` method and instead exposes a `joinWithCode` method, plus all the other inherited (and untouched) methods from the parent class.
+
+While we may have liked to re-use the `join` method name, Dart inheritance rules do not allow doing so.
+
+Then in your client you could move all usages from `client.team` to `client.teamV2`, and could eventually remove the old endpoint. Either by marking it with `@ignoreEndpoint` on the class or deleting it and moving the untouched method implementations you want to keep to the V2 endpoint class.
+
+An alternative pattern to consider would be to move all the business logic for an endpoint into a helper class, and then call into that from the endpoint. In case you want to create a V2 version later, you might be able to reuse most of the underlying business logic through that helper class, and don't have to sub-class the old endpoint. This has the added benefit of the endpoint class clearly listing all exposed methods, and you don't have to wonder what you inherit from the base class.
```
tp, line 198: Personally I think as a general "API evolution" approach the externalized business logic and 2 concrete implementations seem better for the common case.

I think extending endpoints is really more suitable in the module -> app case we discussed, where you do it to minimally modify / extend behavior and not do what otherwise would be breaking changes.

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+  }
+}
+
+class ExcitedGreeter extends GreeterBase {
+  @override
+  Future<String> hello(Session session, String name) async {
+    return '${super.hello(session, name)}!!!';
+  }
+
+  // to expose `wave` on the sub-class, override it like this to drop the `@ignoreEndpoint` annotation
```
dkbast, line 142:
```suggestion
  // to enable `wave` which was marked with `@ignoreEndpoint` on the sub-class, you can override and call the `super` implementation. This will create the method on the endpoint.
```

dkbast: Took me a bit to understand whats going on here 🫣 - but nice that this is now possible :)

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+  @ignoreEndpoint
+  Future<String> wavePerson(Session session, String name) async {
+    throw UnimplementedError();
+  }
+}
+```
+
+In the above example `ExcitedGreeter` inherits from `GreeterBase`. Since the base class is marked as `abstract` no client was generated for it, but now all of its visible endpoint methods would be exposed through the sub-class `ExcitedGreeter`.
+
+The sub-class applies the following modifications though:
```
dkbast, line 158:
```suggestion
The sub-class `ExcitedGreeter` is an example for all possible modifications to the abstract base class
```

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+
+- `hello` is overriden and augments the super class' implementation with a trailing `!!!`
+- `goodbye` is now exposed through the sub-class, as it was not individually hidden and the sub-class does not further augment it
+- `wave`, which is explictly ignored in the base-class is still not exposed, as the sub-class did not opt into re-exposing it
+- `wavePerson` was explicitly overriden to be ignored. Don't worry about the `throw` in the method implementation, which is only needed to satisfy the compiler. In practice this method can not be called from the client anymore.
+
+### API versioning for breaking changes
+
+Endpoint sub-class can be useful when having to do a breaking change on an endpoint, but you want to keep sharing most of it's implementation with the old endpoint.
```
dkbast, line 167:
```suggestion
Endpoint sub-class can be useful when having to do a breaking change on an endpoint, but you need to keep supporting existing clients. Doing so also allows you to keep sharing most of it's implementation with the old endpoint.
```

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+
+```dart
+@Deprecated('Use TeamV2Endpoint instead')
+class TeamEndpoint extends Endpoint {
+  Future<TeamInfo> join(Session session) async { … }
+  
+  // many more methods, like `leave` etc.
+}
+
+class TeamV2Endpoint extends TeamEndpoint {
```
dkbast, line 179: I think we should try to come up with a good naming convention for versioning which is not suffixing numbers - let's maybe discuss this in a bigger round to get more brains involved :)

dkbast (commented):
Thats a pretty sweet new feature :) 🚀 - I think the greeter example is great 👏 - left a few suggestions here and there and would probably want to pass this through grammarly at some point - ping me if you want me to do this

docs/06-concepts/01-working-with-endpoints.md
tp:
```suggestion
  // to enable `wave` which was marked with `@ignoreEndpoint` on the base class, you can override it on the sub-class and call the `super` implementation. This will create the method on the endpoint.
```

In the original suggestion it should be the base class (which had ignored the method). Maybe this is clearer (or too verbose 😅)

docs/06-concepts/01-working-with-endpoints.md
tp:
```suggestion
An endpoint sub-class can be useful when having to do a breaking change on an endpoint, but you need to keep supporting existing clients. Doing so also allows you to keep sharing most of its implementation with the old endpoint.
```

docs/06-concepts/01-working-with-endpoints.md
tp:
```suggestion
The sub-class `ExcitedGreeter` is an example for all possible modifications to the abstract base class and contains the following changes:
```

docs/06-concepts/01-working-with-endpoints.md
tp: Happy to take suggestions. I think while not the nicest (especially since going from "" to "v2"), it is the most common pattern I know of, given that the core entity retains its name.

I think this is also what people would put into their REST API paths, etc.

But you're right, maybe for Serverpod specifically we could do better, e.g. deprecating the old one and thus keeping the default at "team" whereas the old one would become "teamV1" or similar on the client (which the protocol of course staying stable, just the client code-generation behaving in a way so that it looks "evergreen").

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+- `hello` is overriden and augments the super class' implementation with a trailing `!!!`
+- `goodbye` is now exposed through the sub-class, as it was not individually hidden and the sub-class does not further augment it
+- `wave`, which is explicitly ignored in the base-class, is still not exposed, as the sub-class did not opt into re-exposing it
+- `wavePerson` was explicitly overriden to be ignored. Don't worry about the `throw` in the method implementation, which is only needed to satisfy the compiler. In practice, this method can not be called from the client anymore.
+
+### API versioning for breaking changes
+
+An endpoint sub-class can be useful when having to do a breaking change on an endpoint, but you need to keep supporting existing clients. Doing so also allows you to keep sharing most of its implementation with the old endpoint.
+
+Imagine you had a "team" management endpoint where before a user could join if they had an e-mail address ending in the expected domain, but now it should be opened up for anyone to join if they can provide an "invite code".
```
dkbast, line 169:
```suggestion
Imagine you had a "team" management endpoint where before, a user could join if they had an e-mail address ending in the expected domain, but now it should be opened up for anyone to join if they can provide an "invite code".
```

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+    …
+  }
+}
+```
+
+In the above example, we created a new `TeamV2` endpoint, which hides the `join` method and instead exposes a `joinWithCode` method, plus all the other inherited (and untouched) methods from the parent class.
+
+While we may have liked to re-use the `join` method name, Dart inheritance rules do not allow doing so.
+
+In your client, you could then move all usages from `client.team` to `client.teamV2` and eventually remove the old endpoint. Either by marking it with `@ignoreEndpoint` on the class or deleting it and moving the untouched method implementations you want to keep to the V2 endpoint class.
```
dkbast, line 196:
```suggestion
In your client, you could then move all usages from `client.team` to `client.teamV2` and eventually remove the old endpoint. Either mark it with `@ignoreEndpoint` on the class or delete it and move the untouched method implementations you want to keep to the V2 endpoint class.
```

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+
+An alternative pattern to consider would be to move all the business logic for an endpoint into a helper class, and then call into that from the endpoint. In case you want to create a V2 version later, you might be able to reuse most of the underlying business logic through that helper class, and don't have to sub-class the old endpoint. This has the added benefit of the endpoint class clearly listing all exposed methods, and you don't have to wonder what you inherit from the base class.
```
dkbast, line 198:
```suggestion
An alternative pattern to consider would be to move all the business logic for an endpoint into a helper class and then call into that from the endpoint. In case you want to create a V2 version later, you might be able to reuse most of the underlying business logic through that helper class, and don't have to sub-class the old endpoint. This has the added benefit of the endpoint class clearly listing all exposed methods, and you don't have to wonder what you inherit from the base class.
```

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+
+Either approach has its pros and cons, and it depends on the concrete circumstances to pick the one that is most useful. Both give you all the tools you need to extend and update your API while moving clients along gracefully, giving them time to update.
```
dkbast, line 200:
```suggestion
Either approach has pros and cons, and it depends on the concrete circumstances to pick the most useful one. Both give you all the tools you need to extend and update your API while gracefully moving clients along and giving them time to update.
```

docs/06-concepts/01-working-with-endpoints.md
tp: @dkbast: FYI @SandPod just mentioned an idea that `@deprecated` on an endpoint could mean "do not generate a client for this anymore".

That way you could keep the old endpoint protocol alive for a little longer, but get a guarantee that you stopped using it in the next update etc.

That would then also free `client.*` for whatever replacement the developer comes up with.

docs/06-concepts/01-working-with-endpoints.md
tp: Haha, funny moving "gracefully" to "the end" was what it suggested to me earlier :D

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+
+## Endpoint inheritance
+
+Extending existing endpoints from your own app or other modules is possible by subclassing them. If the parent endpoint was marked as `abstract` or `@ignoreEndpoint`, no client code is generated for it, but a client will be generated for your subclass – as long as it does not opt out by any of the aforementioned ways.
+
+```dart
+import 'package:serverpod/serverpod.dart';
+
+abstract class GreeterBase extends Endpoint {
+  Future<String> hello(Session session, String name) async {
```
SandPod, line 118: Suggestion: Break each of these cases out into different sections and target them specifically.

The sections could be:
- - inheritance between concrete classes.
- abstract class with hello endpoint.
- abstract class with require login
- abstract class with ignoreEndpoint
- abstract class with hello endpoint with an override in the concrete class.

And probably in that order. Try to make the examples "real world" examples if possible.

This should make it a bit easier to gauge the capability of inheritance.

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,112 @@ class ExampleEndpoint extends Endpoint {
...
+In the above example, `ExcitedGreeter` inherits from `GreeterBase`. Since the base class is marked as `abstract`, no client was generated for it, but now all of its visible endpoint methods would be exposed through the sub-class `ExcitedGreeter`.
+
+The sub-class `ExcitedGreeter` is an example of all possible modifications to the abstract base class and contains the following changes:
+
+- `hello` is overridden and augments the super class' implementation with a trailing `!!!`
+- `goodbye` is now exposed through the sub-class, as it was not individually hidden, and the sub-class does not further augment it
+- `wave`, which is explicitly ignored in the base-class, is still not exposed, as the sub-class did not opt into re-exposing it
+- `wavePerson` was explicitly overriden to be ignored. Don't worry about the `throw` in the method implementation, which is only needed to satisfy the compiler. In practice, this method can not be called from the client anymore.
+
+### API versioning for breaking changes
```
SandPod, line 165: Suggestion: Move this to the `backwards compatibility` page and integrate it with the text there.

SandPod (changes requested):
Nice foundation!

Please split the documentation up a bit to make it a bit easier to digetst and ensure that we cover all scenarios.

docs/06-concepts/01-working-with-endpoints.md
tp: Before I was worried that going into detail on each would be a bit long (as the current docs seem rather succinct).

But now having changed it, I already like it better, as for each case we can give a clearer explanation and it does not get "blurred" into other behaviors. This should also be much better for sharing, e.g. if one wants to suggest: Hey I think this pattern could be improved by doing what's explained here etc.

SandPod (changes requested):
Nice with clear examples. I think this makes it much easier to digest.

Generally, we try to use the format:
1. Explain what can be done, with a motivation on why is it useful.
2. Give an example
3. Explain the example

Then if there are any additions we add a subheader and go into that specific example.

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,258 @@ class ExampleEndpoint extends Endpoint {
...
+class MyCalculator extends Calculator {
+  Future<int> subtract(Session session, int a, int b) async {
+    return a - b;
+  }
+}
+```
+
+Now the `MyCalculator` endpoint exposes the inherited `add` endpoint method as well as `subtract`.
+
+### Inheriting from an `@ignoreEndpoint` base class
```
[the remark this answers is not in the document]
tp:
```suggestion
### Inheriting from `Endpoint` class marked with `@ignoreEndpoint`
```

Maybe like this?

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,258 @@ class ExampleEndpoint extends Endpoint {
...
+Additionally, further methods could be added to the `MyCalculator` subclass.
+
+### Overriding endpoint methods
+
+Suppose you had a base `Endpoint` class `Greeter` whose behavior you want to adopt.
+
+```dart
+import 'package:serverpod/serverpod.dart';
+
+abstract class GreeterBase extends Endpoint {
```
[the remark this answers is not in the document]
tp: I am for it :)

Let's start that separately and then update the docs in one swoop.

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,258 @@ class ExampleEndpoint extends Endpoint {
...
+    return a - b;
+  }
+}
+```
+
+The generated code will now contain both a `Calculator` and `MyCalculator` client.  
+The `MyCalculator` endpoint will not expose both `add` and `subtract` methods.
+
+To prevent usage of the base `Calculator` endpoint, it might be appropriate to hide it.
+If the base class was defined in another module as shown above, then this is currently not possible. But if it's part of your own code, you could use the `@ignoreEndpoint` annotation from the previous section, or make it `abstract` as shown below.
```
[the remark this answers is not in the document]
tp: The basic idea behind the `abstract` endpoints in the package was that now no split is needed (at least not for the endpoint exposing per se), but the developer still gets the flexibility to opt into or augment the provided endpoint.

But you're right, this probably warrants some explanation to clearly state why you would want to do this.

For "auth 2.0" this gets complicated a bit further as we do still split some packages, but now not in order to give control over the exposed endpoint(s), but due to the dependencies (e.g. `auth_email_account` (the "core") will only depend on the "auth2", whereas `auth_email` (the endpoint) will depend on the default session and user profile implementation, offering a drop-in solution).

But maybe even that is worthwhile information to put in, to explain in which case this might be most useful, and what other aspects to consider.

docs/06-concepts/17-backward-compatibility.md
```diff
@@ -7,3 +7,44 @@ Following a simple set of rules, your server will stay compatible with older app
...
+
+An [endpoint sub-class](/concepts/working-with-endpoints) can be useful when you have to make a breaking change to an entire endpoint but need to keep supporting existing clients. Doing so allows you to share most of its implementation with the old endpoint.
+
+Imagine you had a "team" management endpoint where before a user could join if they had an e-mail address ending in the expected domain, but now it should be opened up for anyone to join if they can provide an "invite code". Additionally, the return type (serialized classes) should be updated across the entire endpoint, which would not be allowed on the existing one.
+
+Transitioning from the current to the new endpoint structure might look like this:
+
+```dart
+@Deprecated('Use TeamV2Endpoint instead')
+class TeamEndpoint extends Endpoint {
```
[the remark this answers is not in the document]
tp: > Or the other way round, hiding the deprecated one in the client (either automatically if its marked deprecated or explicitly like so)

I think hiding it on the client / in the generated code while still exposing it on the server is a great idea. I'll see whether we have a ticket for that already.

> then aliasing the new endpoint

The only issue here is that we can not use the old name on the protocol (as that still needs to support the old methods), but the "Name Override" you suggest could definitely be used on the client (so that the usage stays `client.team.*`, just with the new set of methods).

docs/06-concepts/01-working-with-endpoints.md
SandPod: Rewrite looked good! 👍

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,258 @@ class ExampleEndpoint extends Endpoint {
...
+import 'package:serverpod/serverpod.dart';
+
+class MyCalculator extends Calculator {
+  Future<int> subtract(Session session, int a, int b) async {
+    return a - b;
+  }
+}
+```
+
+The generated code will now contain both a `Calculator` and `MyCalculator` client.
```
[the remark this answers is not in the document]
tp: I adapted the pattern such that the first sentence(s) after each example code (a single one per section now) list the exposed methods.

I did not rewrite it into the `client.x.y` style though, as that seems a little further detached from the example, and also varied the expressions a little bit depending on the focus of the example so that it does not read quite as "mechanical" (as its text, not a table). Let me know what you think :)

docs/06-concepts/01-working-with-endpoints.md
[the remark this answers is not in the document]
tp: Removed, but added a similar explicit warning to the case where one is trying to hide a method from a non-abstract/ignored endpoint, as that still leaves the method externally available by using the super class's endpoint.

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,258 @@ class ExampleEndpoint extends Endpoint {
...
+
+So far, this would not generate any client or server code.
+
+If you then added a subclass that extends this base class, this subclass would be exposed to clients.
+
+```dart
+import 'package:serverpod/serverpod.dart';
+
+class MyCalculator extends Calculator {}
+```
```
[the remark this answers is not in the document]
tp: all examples changed to 1 code snippet per heading.

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,258 @@ class ExampleEndpoint extends Endpoint {
...
+import 'package:serverpod/serverpod.dart';
+
+class MyCalculator extends Calculator {
+  Future<int> subtract(Session session, int a, int b) async {
+    return a - b;
+  }
+}
+```
+
+Now the `MyCalculator` endpoint exposes the inherited `add` endpoint method as well as `subtract`.
```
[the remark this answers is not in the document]
tp: done, added the extra section.

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +89,258 @@ class ExampleEndpoint extends Endpoint {
...
+
+class MyCalculator extends Calculator {}
+```
+
+Now, all of the public endpoint methods of the base class, in this case just `add`, would be exposed by the `MyCalculator` client.
+Additionally, further methods could be added to the `MyCalculator` subclass.
+
+### Overriding endpoint methods
+
+Suppose you had a base `Endpoint` class `Greeter` whose behavior you want to adopt.
```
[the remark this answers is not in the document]
tp: Done in 739337c1fe740d129ff1e91324cdc536974f6930

docs/06-concepts/01-working-with-endpoints.md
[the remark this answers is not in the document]
SandPod: Yeah, that is even better!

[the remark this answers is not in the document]
SandPod: Would be neat to get the headers a bit more similair so that it is easy to pattern match your specific case?

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +91,217 @@ class ExampleEndpoint extends Endpoint {
...
+  @override
+  Future<BigInt> addBig(Session session, BigInt a, BigInt b) async {
+    return super.addBig(session, a, b);
+  }
+}
+```
+
+Since `CalculatorEndpoint` is `abstract` it will not be exposed on the server. `MyCalculatorEndpoint` will expose both the `add` and `addBig` methods, since `addBig` was overridden and thus lost the `@ignoreEndpoint` annotation.
+
+### Building base endpoints for behavior
```
SandPod, line 283: Suggestion(NIT): Should this be at the top? Feels like this might be the most common scenario?

docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -90,3 +91,217 @@ class ExampleEndpoint extends Endpoint {
...
+
+## Endpoint method inheritance
+
+Endpoints can be based on other endpoints using inheritance, like `class ChildEndpoint extends ParentEndpoint`. If the parent endpoint was marked as `abstract` or `@ignoreEndpoint`, no client code is generated for it, but a client will be generated for your subclass – as long as it does not opt out again.
+
+Currently, there are the following possibilities to extend another `Endpoint` class:
+
+### Inheriting from a concrete (visible) `Endpoint` class
+
+Suppose there is an existing `CalculatorEndpoint` (presumably defined in another package, otherwise the existing one could be adapted) which you want to extend with further functionality.
```
SandPod, line 124: Suggestion(NIT - use your judgement here): Could be rephrased as `Given an existing `Endpoint` class, it is possible to extend or modify its behavior without modifying the already exposed methods.

Then we can stay away from the other package talk. That could maybe be mentioned in the introduction?

"Inheritance gives you the possibility to modify the behavior of endpoints defined in other serverpod modules"?

SandPod (approved):
Last two NITs and a comment on a suggestion you already made.

None of them are critical so I trust your judgement there, give me a ping when we should merge it :)

docs/06-concepts/01-working-with-endpoints.md
tp: Not sure, I think it's nice from the original ticket, but not necessarily something most people would do. Maybe a `mixin` would be more appropriate here, as you can really only have 1 parent and thus doing this tiny change to `requireLogin => true` personally feels not like much to intercept the class hierarchy. Maybe this is indeed more useful with scopes? (But then again, the login + admin endpoint is a bit misleading, because it'd suffice to just specify `scopes => {admin}`, as you can not get a scope (with the default tools) without being backed by a logged in user 🤷🏻‍♂️.)

I think the "normal" extension will get more relevant when we start shipping `abstract` endpoints ourselves, in which case this is a concept the developer must be aware of, whereas these non-endpoint method related things are optional.

docs/06-concepts/01-working-with-endpoints.md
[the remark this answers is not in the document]
tp: Aligned the headers for subclassing and method hide/show respectively.

docs/06-concepts/01-working-with-endpoints.md
tp: You're right, there is no reason for that info to live under this heading and not e.g. the `abstract` one etc. 

So I moved it to the higher up introduction as suggested :)

docs/06-concepts/01-working-with-endpoints.md
SandPod: Although another method of implementation might be preferred, determining access rights for a group of endpoints is still a useful feature.

That gives you the possibility to have classes such as `EditRightsEndpoints` or `AuthenticatedEndpoints`.

Not sure where it should live in the documentation, just think the functionality is really useful for code clarity.

SandPod: Keeping as is for now!

PR #272, opened 2025-04-24
feat: Adds global banner for the course waitlist

dkbast:
image

@vlidholt we need to add the link to the form (and double check that the wording is matching)

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -50,11 +50,11 @@ const config = {
         disableSwitch: false,
       },
       announcementBar: {
-        id: 'course_prelaunch',
+        id: 'serverpod_cloud',
         content:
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -49,6 +49,14 @@ const config = {
         respectPrefersColorScheme: true,
         disableSwitch: false,
       },
+      announcementBar: {
+        id: 'course_prelaunch',
+        content:
+          'We are working on an introductory Serverpod course and are looking for your input - if you are interested in early access please  <a target="_blank" rel="noopener noreferrer" href="#">join the waitlist</a>',
```
dkbast, line 55: replace `target="_blank"` with the correct link

dkbast: @vlidholt I think this one is for you

vlidholt (approved):
LGTM 🥳

PR #273, opened 2025-04-28
docs: Update documentation to refer to new `@doNotGenerate` annotation

tp:
instead of deprecated `@ignoreEndpoint`

Based upon https://github.com/serverpod/serverpod/pull/3484

Closes https://github.com/serverpod/serverpod/issues/3478

### DIFF
--- a/docs/06-concepts/01-working-with-endpoints.md
+++ b/docs/06-concepts/01-working-with-endpoints.md
@@ -77,12 +77,12 @@ The return type must be a typed Future. Supported return types are the same as f
 
 ### Ignore an entire `Endpoint` class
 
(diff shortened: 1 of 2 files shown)


docs/06-concepts/01-working-with-endpoints.md
```diff
@@ -77,12 +77,12 @@ The return type must be a typed Future. Supported return types are the same as f
 
 ### Ignore an entire `Endpoint` class
```
tp, line 78: I would leave the headings as "Ignore …" instead of "Do not generate code for an entire `Endpoint` class" etc. as that, to me, succinctly expresses the outcome.

SandPod (approved):
LGTM! 👍 

I trust that you have done a search and replace for all references of `@ignoreEndpoint` in the `docs` directory.

tp:
> I trust that you have done a search and replace for all references of @ignoreEndpoint in the docs directory.

Manually, but yes, I only changed the `docs` files but left the versioned ones untouched:

CleanShot 2025-05-05 at 08 44 11@2x

PR #277, opened 2025-05-13
fix: Add documentation for uuidv7 default option.

SandPod:
Adds documentation for the new UUIDv7 default option.

Screenshot 2025-05-13 at 13 28 35

### DIFF
--- a/docs/06-concepts/02-models.md
+++ b/docs/06-concepts/02-models.md
@@ -195,9 +195,9 @@ This means that `defaultPersist` only comes into play when the model does not pr
 
 #### Boolean
 
-| Type            | Keyword            | Description                                           |
-|-----------------|--------------------|-------------------------------------------------------|
-| **Boolean**     | `true` or `false`  | Sets the field to a boolean value, either `true` or `false`. |
+| Type        | Keyword           | Description                                                  |
+| ----------- | ----------------- | ------------------------------------------------------------ |
+| **Boolean** | `true` or `false` | Sets the field to a boolean value, either `true` or `false`. |
 
 **Example:**
 
@@ -207,10 +207,10 @@ boolDefault: bool, default=true
 
 #### DateTime
 
-| Type                    | Keyword       | Description                                                  |
-|-------------------------|---------------|--------------------------------------------------------------|
-| **Current Date and Time** | `now`        | Sets the field to the current date and time.                  |
-| **Specific UTC DateTime** | UTC DateTime string in the format `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'` | Sets the field to a specific date and time. |
+| Type                      | Keyword                                                          | Description                                  |
+| ------------------------- | ---------------------------------------------------------------- | -------------------------------------------- |
+| **Current Date and Time** | `now`                                                            | Sets the field to the current date and time. |
+| **Specific UTC DateTime** | UTC DateTime string in the format `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'` | Sets the field to a specific date and time.  |
 
 **Example:**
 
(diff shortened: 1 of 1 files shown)


docs/06-concepts/02-models.md
```diff
@@ -306,16 +306,18 @@ stringDefault: String, default='This is a string'
...
-| **UUID String**    | A valid UUID version 4 string | Assigns a specific UUID to the field. |
+| Type              | Keyword                       | Description                                                                                                                                       |
+| ----------------- | ----------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Random UUID**   | `random`                      | Generates a random UUID. On the Dart side, `Uuid().v4obj()` is used. On the database side, `gen_random_uuid()` is used.                           |
+| **Random UUIDv7** | `random_v7`                   | Generates a random UUIDv7. On the Dart side, `Uuid().v7obj()` is used. On the database side, a generated `gen_random_uuid_v7()` function is used. |
+| **UUID String**   | A valid UUID version 4 string | Assigns a specific UUID to the field.                                                                                                             |
```
dkbast, line 313: ⛏ As a user I would like to know the format:

'xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx' where M is the UUID version field. The upper two or three bits of digit N encode the variant. E.g. '550e8400-e29b-41d4-a716-446655440000'

docs/06-concepts/02-models.md
```diff
@@ -306,16 +306,18 @@ stringDefault: String, default='This is a string'
...
 
 **Example:**
 
 ```yaml
 uuidDefaultRandom: UuidValue, default=random
 uuidDefaultUuid: UuidValue, default='550e8400-e29b-41d4-a716-446655440000'
+uuidDefaultRandomUuidV7: UuidValue, default=random_v7
```
dkbast, line 320: If possible, it would be great to have "speaking names" here and add ### comments for additional context

dkbast (approved):
Left two ideas for improvements - but feel free to treat them as optional

docs/06-concepts/02-models.md
SandPod: Could you give me an example of a "speaking" name? :)

```
uuidDefaultRandomV4
uuidDefaultFixedValue
uuidDefaultRandomV7
```
?

docs/06-concepts/02-models.md
SandPod: Fixed: [e63a0ba](https://github.com/serverpod/serverpod_docs/pull/277/commits/e63a0bae34e3d8678eab603beff2dcba966eaf21)

docs/06-concepts/02-models.md
dkbast: This is what ChatGPT came up with:
```yaml
### Information about a company.
class: Company
fields:
  ### Primary identifier for the company – we use a time-sortable
  ### UUIDv7 so that records created close in time cluster together
  ### in the database and in logs.
  companyId: UuidValue, default=random_v7

  ### Public API token that the company includes when
  ### calling our endpoints. A plain random UUID is perfect here
  ### because it never needs ordering semantics – it just
  ### needs to be unpredictable.
  apiToken: UuidValue, default=random

  ### ID that links back to the legacy CRM.  We hard-code it so
  ### that migrated data keeps the *same* identifier everywhere.
  legacyCrmId: UuidValue, default='550e8400-e29b-41d4-a716-446655440000'

  ### The name of the company.
  name: String

  ### The date the company was founded, if known.
  foundedDate: DateTime?

  ### A list of people currently employed at the company.
  employees: List<Employee>
```

SandPod: Aah yeah, that makes sense. But perhaps we should rework the whole page in that case? Since we are only using "named" fields that represent the value in the other spots currently.

Let's revisit this when updating the "working wth models" page as it feels like a general style improvement.

PR #278, opened 2025-05-13
fix: Add documentation for concurrent future calls

SandPod:
Adds documentation for the new configuration option on future calls.

### Configuration page
Screenshot 2025-05-13 at 13 59 43

---- 
Screenshot 2025-05-13 at 14 00 11
----

### Schedule page

Screenshot 2025-05-13 at 14 06 10

----

Screenshot 2025-05-13 at 14 05 54


dkbast (approved):
LGTM 🚀

docs/06-concepts/07-configuration.md
```diff
@@ -120,6 +123,12 @@ maxRequestSize: 524288
 sessionLogs:
   persistentEnabled: true
   consoleEnabled: true
+
+futureCallExecutionEnabled: true
```
dkbast, line 127: Should we add a warning to our LSP if this is false but there are values set for futureCall ?

SandPod: I think it would make sense to give a "hint" that this is the case. But it could actually be used to just disable future calls without changing the rest of the configuration.

But our LSP currently doesn't analyze the configuration files. That is something we are aiminig to introduce still.

docs/06-concepts/14-scheduling.md
```diff
@@ -81,3 +83,51 @@ This identifier can then be used to cancel all future calls registered with said
...
+```
+
+### Enable or disable future call execution
+
+This option allows you to enable or disable the execution of future calls. By default, it is set to `true`. You might want to disable future call execution in environments where you don't want background tasks to run, such as during testing or in a staging environment where you want to focus on API behavior without triggering scheduled tasks.
+
+Example configuration:
+
+```yaml
+futureCallExecutionEnabled: true
```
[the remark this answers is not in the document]
SandPod: Good catch! Will fix! 👍

PR #279, opened 2025-05-13
docs: Update documentation to reflect that `AuthenticationInfo` now stores a `String` `userIdentifier`

tp:


docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -25,7 +25,11 @@ final pod = Serverpod(
 );
 ```
 
-In the above example, the `authenticationHandler` callback is overridden with a custom validation method. The method returns an `AuthenticationInfo` object with user id `1` and no scopes if the token is valid, otherwise `null`.
+In the above example, the `authenticationHandler` callback is overridden with a custom validation method. The method returns an `AuthenticationInfo` object with `userIdentifier` `"1"` and no scopes if the token is the literal "valid", otherwise `null`.
+
```
SandPod, line 29: Suggestion: Update the example to use the String notation instead, then we can explain the "legacy" userId method below in the note along that the `userIdentifier` takes and Object that `toString` is called on and that the internal representation is a `String` that is accessible through the `user` field.

docs/06-concepts/11-authentication/05-custom-overrides.md
```diff
@@ -25,7 +25,11 @@ final pod = Serverpod(
...
+:::note
+The `userIdentifier` passed to the `AuthenticationInfo` constructor, as the first parameter, will always be converted to a `String` and thus stored internally. For historical reasons there is a `userId` getter on the object that returns an `int`, if the string can be converted.
```
SandPod, line 31: Question: A reference to the `userId` field is also used in the `02-basics.md` page. Should we update that section to to reflect that the "String" is available along with the parsing of the `userId`?

SandPod (commented):
Two suggestions/Questions.

docs/06-concepts/11-authentication/05-custom-overrides.md
tp: But wouldn't it read unusual to pass a literal number as `String` (like `"1"`).

I think in the current setup, with `1` passed, but the text mentioning `"1"` it's explicit that the conversion took place.

When a string is already passed in, there need not be any expectation of the "stringification".

docs/06-concepts/11-authentication/05-custom-overrides.md
tp: Since the initial examples don't do anything specific with it (e.g. pass it to an old API requiring) `int` I updated them to use `userIdentifier`.

But we still need the `int` `userId` for the later example snippet that calls into `serverpod_auth` and presumably runs in the same method.

docs/06-concepts/11-authentication/02-basics.md
```diff
@@ -2,11 +2,12 @@
 
 Serverpod automatically checks if the user is logged in and if the user has the right privileges to access the endpoint. When using the `serverpod_auth` module you will not have to worry about keeping track of tokens, refreshing them or, even including them in requests as this all happens automatically under the hood.
 
-The `Session` object provides information about the current user. A unique `userId` identifies a user. You should use this id whenever you a referring to a user. Access the id of a signed-in user through the `authenticated` asynchronous getter of the `Session` object.
+The `Session` object provides information about the current user. A unique `userIdentifier` identifies a user. You should use this id whenever you a referring to a user. Access the id of a signed-in user through the `authenticated` asynchronous getter of the `Session` object. Since the default implementation of `serverpod_auth` uses numeric IDs for the users, there is a convenience getter `userId` on the `AuthenticationInfo`, which returns the integer value.
```
tp, line 5:
```suggestion
The `Session` object provides information about the current user. A unique `userIdentifier` identifies a user. You should use this id whenever you a referring to a user. Access the id of a signed-in user through the `authenticated` asynchronous getter of the `Session` object. For historical reasons, there is a `userId` getter on the object that returns an `int`, if the string can be converted.
```

docs/06-concepts/11-authentication/05-custom-overrides.md
tp, line 31:
```suggestion
The `userIdentifier` passed to the `AuthenticationInfo` constructor, as the first parameter, will always be converted to a `String` and thus stored internally. Since the default implementation of `serverpod_auth` uses numeric IDs for the users, there is a convenience getter `userId`, which returns the integer value.
```

SandPod (approved):
LGTM! 👍

PR #288, opened 2025-06-03
fix: Remove all pre stable release versions of the documentation.

SandPod:
Removes all versions of the documentation made before 1.0.0.


dkbast (approved):
Nice cleanup :)

PR #289, opened 2025-06-03
fix: Update code examples to include session parameter

dkbast:
The code examples did not include the session parameter


SandPod (approved):
LGTM! Could you apply these fixes to the already released documentation as well?

And since we are just cutting a new version of the documentation. Let's coordinate this so that the changes are included there.

PR #290, opened 2025-06-03
chore: Makes it easier to discover joins

dkbast:


SandPod (approved):
LGTM! 👍 

Please apply to all versions of the documentation where it is applicable.

PR #291, opened 2025-06-03
docs: Add full rollback example

dkbast:
Adds the full instructions on how to do the rollback - if someone does not read the whole page but jumps to the section in the navigation they will not necessarily know where to look


SandPod (approved):
LGTM! 👍 Please apply to all applicable versions of the documentation.

PR #300, opened 2025-06-17
feat: Adds a layer of polish to the documentation.

vlidholt:


SandPod (approved):
Move fast and break things.

Do we wan't to apply these changes to older documentations or are we only focusing on improving the newest version of the documentation?

PR #301, opened 2025-06-23
docs: Update environment variable naming conventions in configuration documentation

exaby73:
Since we have a new way of defining secrets through env (https://github.com/serverpod/serverpod/pull/3656), we should recommend users use this way to define all secrets instead of having special env vars for built-in secrets. This ensures consistency across all secrets, built-in and custom

## Summary by CodeRabbit

- **Documentation**
  - Updated documentation to reflect new naming conventions for environment variables related to secrets, now using the prefix SERVERPOD_PASSWORD_ followed by the key name in camelCase. This affects variables for database, service, Redis, HMAC, and AWS keys. Descriptions and default values remain unchanged.


SandPod (approved):
LGTM! 👍

PR #311, opened 2025-07-09
docs: Document generator.yaml

exaby73:


tp (approved):
backfill looks good to me as well

PR #312, opened 2025-07-09
docs: Add upgrade guide for transitioning from Widget to Component in Serverpod 3.0

exaby73:


docs/08-upgrading/06-upgrade-to-three.md
```diff
@@ -0,0 +1,226 @@
...
+| `WidgetRoute`    | `ComponentRoute`    |
+| `WidgetJson`     | `JsonComponent`     |
+| `WidgetRedirect` | `RedirectComponent` |
+| `WidgetList`     | `ListComponent`     |
+
+If you're importing from the relic library, the import path has changed:
+
+```dart
+// Before
+import 'package:serverpod/src/relic/web_widget.dart';
```
tp, line 22: Would any consumer really use these `src/` paths?

If so, I think we should add a proper export, as most projects have probably a lint against src imports.

tp (approved):

PR #313, opened 2025-07-16
docs: Document the Session API

exaby73:
Closes https://github.com/serverpod/serverpod/issues/3161 , https://github.com/serverpod/serverpod/issues/3419

### DIFF
--- a/docs/06-concepts/05-sessions.md
+++ b/docs/06-concepts/05-sessions.md
@@ -2,6 +2,12 @@
 
 The `Session` object is the central context for server operations in Serverpod. It provides access to the database, caching, authentication, file storage, messaging, and request information during method calls and streaming connections.
 
+:::note
+
+Don't confuse Session objects with user sessions. Session objects are request-scoped server contexts that exist only during a request, while user sessions are persistent authentication states. See the [Authentication documentation](./11-authentication/01-setup.md) for managing user sessions.
+
+:::
+
 ## Core properties
 
 Every `Session` provides access to these essential properties:
--- a/docs/06-concepts/05-sessions.md
+++ b/docs/06-concepts/05-sessions.md
@@ -46,57 +46,118 @@ For authentication features like user management and auth providers, see the [Au
 
 ## Session types
 
-Serverpod creates different session types based on the context:
+Serverpod automatically creates different session types based on the request context and endpoint configuration:
 
 ### MethodCallSession
 
-Created for endpoint method calls. Provides access to HTTP request data:
+Created when a client makes a standard HTTP request to an endpoint method that returns `Future<T>`. This is the most common session type for REST-like API calls.
 
 ```dart
-if (session is MethodCallSession) {
-  Uri uri = session.uri;
-  String body = session.body;
-  Map<String, dynamic> queryParams = session.queryParameters;
-  Request httpRequest = session.request;
-  
-  // Access client information
-  String remoteInfo = session.remoteInfo ?? 'unknown';
+// This endpoint creates a MethodCallSession
+Future<String> hello(Session session, String name) async {
+  if (session is MethodCallSession) {
+    // Access HTTP request details
+    Uri uri = session.uri;
+    String body = session.body;
+    Map<String, dynamic> queryParams = session.queryParameters;
+    Request httpRequest = session.request;
+    
+    // Access client information
+    String remoteInfo = session.remoteInfo ?? 'unknown';
+  }
+  return 'Hello $name';
 }
 ```
 
 ### MethodStreamSession
 
-Created for streaming endpoint connections with a unique connection ID:
+Created when an endpoint method returns `Stream<T>` or has stream parameters. The framework establishes a WebSocket connection and maintains it for the duration of the stream.
 
 ```dart
-if (session is MethodStreamSession) {
-  UuidValue connectionId = session.connectionId;
+// This endpoint creates a MethodStreamSession
+Stream<int> countToTen(Session session) async* {
+  if (session is MethodStreamSession) {
+    // Access the unique connection ID
+    UuidValue connectionId = session.connectionId;
+  }
+  
+  for (var i = 1; i <= 10; i++) {
+    yield i;
+    await Future.delayed(Duration(seconds: 1));
+  }
 }
 ```
 
 ### FutureCallSession
 
-Created when executing scheduled future calls:
+Created automatically by Serverpod when executing scheduled tasks. These sessions run in the background without any client connection.
 
 ```dart
-if (session is FutureCallSession) {
-  String futureCallName = session.futureCallName;
+// Schedule a task (this doesn't create a session yet)
+await serverpod.futureCallWithDelay(
+  'sendEmail',
+  EmailData(to: 'user@example.com'),
+  Duration(hours: 1),
+);
+
+// When the task executes, Serverpod creates a FutureCallSession
+class SendEmailCall extends FutureCall {
+  @override
+  Future<void> invoke(Session session, SerializableModel? object) async {
+    if (session is FutureCallSession) {
+      String taskName = session.futureCallName; // 'sendEmail'
+    }
+    // Send the email
+  }
 }
 ```
 
 ### WebCallSession
 
-Created for web server requests when handling custom web routes:
+Created when the web server handles HTTP requests to custom web routes (not API endpoints). Used for serving web pages, handling form submissions, and other web-specific operations.
 
 ```dart
-if (session is WebCallSession) {
-  // Handle web-specific operations
+// Configure a web route
+webServer.addRoute(
+  RouteRoot(
+    path: '/hello',
+    widget: MyWidget(),
+  ),
+  'hello',
+);
+
+// When someone visits /hello, a WebCallSession is created
+class MyWidget extends Component {
+  @override
+  Future<Component> build(Session session) async {
+    if (session is WebCallSession) {
+      // Handle web-specific operations
+      // Authentication is typically handled via cookies
+    }
+    return Text('Hello Web!');
+  }
 }
 ```
 
 ### InternalSession
 
-Used internally by Serverpod for operations outside client calls.
+Created for internal server operations that aren't triggered by client requests. Serverpod creates one persistent InternalSession at startup, and you can create additional ones for background tasks.
+
+```dart
+// Serverpod creates this automatically at startup
+// Used for system operations, migrations, etc.
+
+// You can create additional InternalSessions
+var session = await Serverpod.instance.createSession(
+  enableLogging: false,
+);
+try {
+  // Perform background operations
+  await User.db.deleteWhere(session, where: (t) => t.inactive);
+} finally {
+  await session.close();
+}
+```
 
 ## Working with sessions
 
--- a/docs/06-concepts/05-sessions.md
+++ b/docs/06-concepts/05-sessions.md
@@ -17,8 +17,6 @@ Every `Session` provides access to these essential properties:
 - **`storage`**: Cloud storage operations for files. See [File uploads documentation](./12-file-uploads.md)
 - **`messages`**: Real-time messaging between sessions. See [Streams documentation](./15-streams.md)
 - **`passwords`**: Passwords loaded from `config/passwords.yaml` and environment variables. See [Configuration documentation](./07-configuration.md)
-- **`server`**: The Server instance that created this session
-- **`serverpod`**: Access to the Serverpod singleton via `server.serverpod`
 
 ## Session identification
 
--- a/docs/06-concepts/05-sessions.md
+++ b/docs/06-concepts/05-sessions.md
@@ -12,11 +12,11 @@ Don't confuse Session objects with user sessions. Session objects are request-sc
 
 Every `Session` provides access to these essential properties:
 
-- **`db`**: Database access for queries and transactions
-- **`caches`**: Access to local and distributed caches
-- **`storage`**: Cloud storage operations for files
-- **`messages`**: Real-time messaging between sessions
-- **`passwords`**: Passwords loaded from `config/passwords.yaml`
+- **`db`**: Database access for queries and transactions. See [Database documentation](./06-database/01-connection.md)
+- **`caches`**: Access to local and distributed caches. See [Caching documentation](./08-caching.md)
+- **`storage`**: Cloud storage operations for files. See [File uploads documentation](./12-file-uploads.md)
+- **`messages`**: Real-time messaging between sessions. See [Streams documentation](./15-streams.md)
+- **`passwords`**: Passwords loaded from `config/passwords.yaml` and environment variables. See [Configuration documentation](./07-configuration.md)
 - **`server`**: The Server instance that created this session
 - **`serverpod`**: Access to the Serverpod singleton via `server.serverpod`
 
--- a/docs/06-concepts/05-sessions.md
+++ b/docs/06-concepts/05-sessions.md
@@ -10,202 +10,65 @@ A Serverpod Session should not be confused with the concept of "web sessions" or
 
 :::
 
+## Quick reference
+
+### Essential properties
+
+- **`db`** - Database access. [See database docs](./06-database/01-connection.md)
+- **`caches`** - Local and distributed caching. [See caching docs](./08-caching.md)
+- **`storage`** - File storage operations. [See file uploads](./12-file-uploads.md)
+- **`messages`** - Server events for real-time communication. [See server events docs](./16-server-events.md)
+- **`passwords`** - Credentials from config and environment. [See configuration](./07-configuration.md)
+
+### Key methods
+
+- **`log(message, level)`** - Add log entry
+- **`close(error, stackTrace)`** - Manual session cleanup
+- **`addWillCloseListener(callback)`** - Register cleanup callback
+
 ## Session types
 
-### Automatic sessions
+Serverpod creates different session types based on the context:
 
-These sessions are automatically managed by Serverpod - created when a request arrives and closed when the request completes:
+| Type | Created For | Lifetime | Common Use Cases |
+|------|-------------|----------|------------------|
+| **MethodCallSession** | `Future<T>` endpoint methods | Single request | API calls, CRUD operations |
+| **WebCallSession** | Web server routes | Single request | Web pages, form submissions |
+| **MethodStreamSession** | `Stream<T>` endpoint methods | Stream duration | Real-time updates, chat |
+| **StreamingSession** | WebSocket connections | Connection duration | Live dashboards, multiplayer |
+| **FutureCallSession** | Scheduled tasks | Task execution | Email sending, batch jobs |
+| **InternalSession** | Manual creation | Until closed | Background work, migrations |
 
-#### MethodCallSession - Standard API calls
-
-Created for: `Future<T>` endpoint methods  
-Lifetime: Single request-response cycle  
-Common in: Data fetching, CRUD operations, business logic
+### Example: Automatic session (MethodCallSession)
 
 ```dart
 Future<String> hello(Session session, String name) async {
-  // You're working with a MethodCallSession here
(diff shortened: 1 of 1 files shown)


docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,292 @@
...
+}
+```
+
+### StreamingSession
+
+Created for WebSocket connections. Allows dynamic endpoint updates:
+
+```dart
+if (session is StreamingSession) {
+  session.updateAuthenticationKey(newKey);
```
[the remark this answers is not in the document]
Isakdl: Indeed and StreamingSession is the old streaming interface that is deprecated.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,292 @@
...
+  );
+}
+```
+
+### Session lifecycle
+
+Sessions can have listeners for cleanup operations:
+
+```dart
+// Add a listener that runs when session closes
```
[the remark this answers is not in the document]
Isakdl: Agreed, I think it is important to discuss the life cycle of a session in this doc and explain when different calls happen. Putting these listeners into that context. Also the example of remove listener just sends a random variable into the method without any explanation. Instead explicitly show how you create that variable and then how you use it to remove the listener.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,292 @@
...
+```dart
+// Listen for messages
+session.messages.addListener('chat', (message) {
+  print('Received: $message');
+});
+
+// Send a message to local server
+await session.messages.postMessage('chat', MyMessage());
+
+// Send globally across cluster
```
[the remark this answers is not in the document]
Isakdl: It does yes

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,292 @@
...
+Every `Session` provides access to these essential properties:
 
-You can use the Session object to access the IP address of the client calling a method. Serverpod includes an extension on `HttpRequest` that allows you to access the IP address even if your server is running behind a load balancer.
+- **`db`**: Database access for queries and transactions
+- **`caches`**: Access to local and distributed caches
+- **`storage`**: Cloud storage operations for files
+- **`messages`**: Real-time messaging between sessions
+- **`passwords`**: Passwords loaded from `config/passwords.yaml`
+- **`server`**: The Server instance that created this session
+- **`serverpod`**: Access to the Serverpod singleton via `server.serverpod`
```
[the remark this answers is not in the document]
exaby73: Both point to the same object. `_innerInitializeServerpod` set's the static `_instance` to `this`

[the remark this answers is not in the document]
Isakdl: Honestly, I don't think we should expose `serverpod` or `server` on the session, that is a different question though.  But not sure if it makes sense to document it? Maybe?

Isakdl (commented):
Looking at this I'm a little skeptical on the content / structure. We are describing very mechanical surface level content, most of which is described in other parts of the docs already. It might make sense to include it, but how does a user navigate this information? If I want to work with file upload am I expected to go to this page? We don't have backlinks to the content where these features are described in detail.

I think one of the more important things to describe is the lifecycle of the session, when is it created and why? When is it destroyed and why? What happens if you try to use a session when it has been disposed of? Show an example of such invalid use of a session, e.g. trying to use a session after your endpoint has returned.

How should a user get around this? E.g. either schedule a future call or create an internal session. Describe how to manage the lifecycle of an internal session as the user is responsible for calling `close`. 

When and how is the authentication field populated.

How does the logging system work together with the session, how are logs written, what are the lifecycle of the logs? For example logs are written after the close of a session for method calls, but continuously written for streams (this can be changed with a setting though!). 

As I mentioned before I'm a little skeptical on showing the different Session types we have. It is true that we have them and that you can use them in theory, however I don't think it is a great interface having to do type conversions in your code. Long term I think the base session should include all necessary information. Does it make sense to document the types today anyway? Maybe...

Anyway I think this documentation page should focus more on life-cycle and inner workings of things rather than just describing the methods that are available. Not wrong to describe them too, but I think that has way less value than describing mechanics that you cannot discover by simply navigating the code in your IDE.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,476 @@
...
+```dart
+class EmailSender extends FutureCall {
+  @override
+  Future<void> invoke(Session session, EmailData? data) async {
+    // FutureCallSession created just for this task
+    await sendEmail(session, data!);
+    // Closes when task completes
+  }
+}
+```
```
Isakdl, line 339: Comment: Not sure about this section. The problem I see is that it is not clear to me what is special about the session in each case, we are just using the normal Session interface. Which I btw think is good, but that kind of makes the value of this text questionable. 

No actionable feedback, I'm not sure what a better alternative is..

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,476 @@
...
+  var service = await ExternalService.connect();
+
+  // Ensure cleanup even if session closes unexpectedly
+  session.addWillCloseListener((session) async {
+    await service.disconnect();
+  });
+
+  // Use service...
+}
+```
```
Isakdl, line 428: Suggestion: This should be described in a section around the lifecycle rather than under best practices as this is a core concept not a style preference.

Isakdl (changes requested):
Much better content!

I have made several suggestions, the most important one around the flow chart. I think this will bring a lot of clarity to the document. 

There is also a question around order and structure that could have an overview, but I think that might become clearer when the flowchart is in place!

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+- Initializes resource accessors (database, cache, storage, messaging)
+- Stores the authentication key (if provided in the request)
+
+Sessions are normally created automatically by the framework. The exception is `InternalSession`, which you create manually for background tasks (see [InternalSession - System operations](#internalsession---system-operations)).
+
+### Execute Endpoint Method
+
+During method execution, the session provides access to all server resources:
+
+1. **Authentication** - Validated automatically before execution if endpoint has `requireLogin: true` or `requiredScopes`, otherwise lazy-loaded on first access to `session.authenticated`
```
SandPod, line 63:
```suggestion
1. **Authentication** - Resolved before execution if endpoint has `requireLogin: true` or `requiredScopes`, otherwise lazy-loaded on first access to `session.authenticated`
```

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+2. **Logging** - Log entries accumulate in memory during execution
```
SandPod, line 64: Discussion: This is only true for non-streaming endpoints. It might be better to document in connection to those?

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+```
+
+### Close Session
+
+Sessions close either automatically or manually:
+
+**Automatic closure:**
+
+- Method calls: After the endpoint method returns
+- Streaming: When the WebSocket connection ends
```
SandPod, line 92: Discussion: There is a difference between streaming endpoints and streaming methods here.

A streaming method will close it's session once the method is complete. A streaming endpoint shares the session between all streams in the websocket and is closed when the websocket is closed.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+
+**Manual closure:**
 
-You can use the Session object to access the IP address of the client calling a method. Serverpod includes an extension on `HttpRequest` that allows you to access the IP address even if your server is running behind a load balancer.
+- Internal sessions: Must call `session.close()` explicitly
+
+When closing, the session:
+
+1. Becomes unusable (further operations throw `StateError`)
+2. Executes any cleanup callbacks registered with `addWillCloseListener()`
```
SandPod, line 102: Note: There is an issue for documenting this that could be closed when we document this: https://github.com/serverpod/serverpod/issues/3419

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+  session.addWillCloseListener((session) async {
+    await service.disconnect();
+  });
+
+  // Use service...
+}
 ```
 
-:::
+### Write logs to database
```
SandPod, line 122: Question/Discussion: Shouldnt this be part of the `Logging mechanics` part of the documentation?

The header structure of this page feels a bit cluttered, I think we can provide a bit more guidance and overview to the reader by splitting up the details from a clear overview to start with. To me, how logs are written feels more like a "dig down" detail rather than overview picture.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+After the session closes, all accumulated logs are written to the database in a single batch:
 
-In most cases, Serverpod manages the life cycle of the Session objects for you. A session is created for a call or a streaming connection and is disposed of when the call has been completed. In rare cases, you may want to create a session manually. For instance, if you are making a database call outside the scope of a method or a future call. In these cases, you can create a new session with the `createSession` method of the `Serverpod` singleton. You can access the singleton by the static `Serverpod.instance` field. If you create a new session, you are also responsible for closing it using the `session.close()` method.
+- All log entries stored in memory during execution are persisted
+- Database query timings and counts are recorded
+- Session metadata (duration, endpoint, method) is saved
+- Returns a session log ID for debugging and tracing
 
-:::note
+**Exception:** Streaming sessions write logs immediately instead of batching.
```
SandPod, line 131: Suggestion: This should be part of the normal section.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+  });
+
+  // Use service...
+}
 ```
 
-:::
+### Write logs to database
 
-## Creating sessions
```
SandPod, line 18: Suggestion: Retain this section to document intenal session and focus on only the automatically created sessions in diagrams and the documentation.

This feels like a deeper detail that might be best explained in it's own section since the user is then responsible for the lifecycle of the section.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+
+  return data;
+}
+```
+
+## Authentication flow
+
+Authentication in sessions works differently based on endpoint configuration:
+
+### For endpoints with `requireLogin: true` or `requiredScopes`
```
SandPod, line 160: Discussion: These headers are not really helpful in providing an overview of the page. Perhaps Authentication should be a subsection to Session lifecycle and that would be enough?

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+```
+
+## Quick reference
+
+### Essential properties
+
+- **`db`** - Database access. [See database docs](./06-database/01-connection.md)
+- **`caches`** - Local and distributed caching. [See caching docs](./08-caching.md)
+- **`storage`** - File storage operations. [See file uploads](./12-file-uploads.md)
+- **`messages`** - Inter-session messaging. [See streams docs](./15-streams.md)
```
SandPod, line 445: Suggestion: Change to `Server events` see server events section in documentation: https://docs.serverpod.dev/concepts/server-events

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+- **`passwords`** - Credentials from config and environment. [See configuration](./07-configuration.md)
+
+### Key methods
+
+- **`authenticated`** - Get authentication info (pre-validated for protected endpoints, lazy-loaded for others)
+- **`isUserSignedIn`** - Quick auth check
```
SandPod, line 451: Discussion: Not sure if this should be promoted?

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,467 @@
...
+- **`log(message, level)`** - Add log entry
+- **`close(error, stackTrace)`** - Manual session cleanup
+- **`addWillCloseListener(callback)`** - Register cleanup callback
+
+### Testing
+
+Use `TestSessionBuilder` in integration tests. [See testing docs](./19-testing/01-get-started.md)
```
SandPod, line 458: Suggestion: Explain what the purpose would be to use the test session builder.

What is the purpose of having this section here? Please make it clear why the user might want to use the test session builder and read more details around it.

SandPod (changes requested):
I visual representation of the lifecycle of the session for sure helps.

I still feel like the page lacks a story telling structure and does not provide an adequate overview to easily get a "get started" understanding of what the session is or how it works.

I think this can be improved by thinking about the header structure and the story telling of the session.

Perhaps we can take a call to discuss the style to use here?

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,254 @@
...
+:::
+
+## Quick reference
+
+### Essential properties
+
+- **`db`** - Database access. [See database docs](./06-database/01-connection.md)
+- **`caches`** - Local and distributed caching. [See caching docs](./08-caching.md)
+- **`storage`** - File storage operations. [See file uploads](./12-file-uploads.md)
+- **`messages`** - Server events for real-time communication. [See server events docs](./16-server-events.md)
```
SandPod, line 20: To use the same language as the Server events documentation and make it clera that it is used both within and across multiple servers.

```suggestion
- **`messages`** - Server events for real-time communication within and across servers. [See server events docs](./16-server-events.md)
```

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,254 @@
...
+- **`passwords`** - Credentials from config and environment. [See configuration](./07-configuration.md)
+
+### Key methods
+
+- **`log(message, level)`** - Add log entry
+- **`close(error, stackTrace)`** - Manual session cleanup
```
SandPod, line 26: Suggestion: Remove, this is only used in an advanced use-case and should not be used in the general case.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,254 @@
...
+| **StreamingSession** | WebSocket connections | Connection duration | Live dashboards, multiplayer |
+| **FutureCallSession** | Scheduled tasks | Task execution | Email sending, batch jobs |
+| **InternalSession** | Manual creation | Until closed | Background work, migrations |
+
+### Example: Automatic session (MethodCallSession)
 
 ```dart
-session as MethodCallSession;
-var ipAddress = session.httpRequest.remoteIpAddress;
+Future<String> hello(Session session, String name) async {
```
SandPod, line 45: Suggestion: Include the endpoint definition in the example, otherwise it is not clear that this is an endpoint.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,254 @@
...
-:::note
+**Important**: Always use try-finally with InternalSession to prevent memory leaks.
 
-It's not recommended to keep a session open indefinitely as it can lead to memory leaks, as the session stores logs until it is closed. It's inexpensive to create a new session, so keep them short.
+## Session lifecycle
 
-:::
+```mermaid
+flowchart TB
+    Request([Client Request]) --> Create[Session Created]
```
SandPod, line 75: Suggestion: This step could also be a future call that is triggered. Documentation should reflect that, could we use a more general term instead of client request?

Since you mention `Execute Endpoint Method`, perhaps we should make it clear that we are describing an endpoint method?

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,254 @@
...
+    Init --> Active[Execute Endpoint Method]
+    Active --> Close[Close Session]
+    Close --> End([Request Complete])
+```
+
+Sessions follow a predictable lifecycle from creation to cleanup. When a client makes a request, Serverpod automatically creates the appropriate session type (see table above), initializes it with a unique ID, and sets up access to resources like the database, cache, and file storage. The only exception is `InternalSession`, which you create manually for background tasks.
+
+During the active phase, your endpoint method executes with full access to Serverpod resources through the session. You can query the database, write logs, send messages, and access storage - all operations are tracked and tied to this specific session. When the method completes, most sessions close automatically, writing any accumulated logs to the database and releasing all resources.
+
+Manual sessions (`InternalSession`) require explicit closure with `session.close()`. Forgetting to close these sessions causes memory leaks as logs accumulate indefinitely. Always use try-finally blocks to ensure proper cleanup. After any session closes, attempting to use it throws a `StateError`.
```
SandPod, line 86: Suggestion: Break this out to a separate header, under `Session lifecycle`, and include the part in the first paragraph as 

Perhaps you could use `Internal Sessions` as the title?

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,254 @@
...
+
+## Logging
+
+Serverpod batches log entries for performance. During normal operations, logs accumulate in memory and are written to the database in a single batch when the session closes. This includes all your `session.log()` calls, database query timings, and session metadata. The exception is streaming sessions, which write logs continuously by default to avoid memory buildup during long connections. **Warning**: If you forget to close a manual session (`InternalSession`), logs remain in memory indefinitely and are never persisted - this is a common cause of both memory leaks and missing debug information.
```
SandPod, line 90: Suggestion: Transform the `warning` into a warning box.

SandPod, line 90: Suggestion: Mention the specific sessions that write logs continuously by type. `MethodStreamSession` and `StreamingSession`.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,254 @@
...
+    // Handle error appropriately
+  } finally {
+    await session.close();
+  }
+}
+```
+
+## Testing
+
+When testing endpoints, you need a way to create valid sessions without going through the full HTTP request cycle. The `TestSessionBuilder` provides a controlled environment for integration tests:
```
SandPod, line 243: And update the example so that a we switch out some form of property for the user, for example

```
var authenticatedSessionBuilder = sessionBuilder.copyWith(
      authentication:
          AuthenticationOverride.authenticationInfo(userId, {Scope('user')}),
    );
```

```suggestion
When testing endpoints, the `TestSessionBuilder` can be used to simulate and configure session properties for controlled test scenarios:
```

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,254 @@
...
-:::tip
+:::note
+
+A Serverpod Session should not be confused with the concept of "web sessions" or "user sessions" which persist over multiple API calls. See the [Authentication documentation](./11-authentication/01-setup.md) for managing persistent authentication.
+
+:::
+
+## Quick reference
+
+### Essential properties
```
SandPod, line 15: Suggestion: Missing authentication info, that is a major component of the session that should be covered somewhere.

docs/06-concepts/05-sessions.md
```diff
@@ -1,26 +1,254 @@
...
+    await session.close(); // Must close manually!
+  }
+}
+```
 
-:::note
+**Important**: Always use try-finally with InternalSession to prevent memory leaks.
 
-It's not recommended to keep a session open indefinitely as it can lead to memory leaks, as the session stores logs until it is closed. It's inexpensive to create a new session, so keep them short.
+## Session lifecycle
```
SandPod, line 71: Suggestion: Missing documentation for `addWillCloseListener`. Please add documentation for this under this section.

SandPod (commented):
I still don't think the documentation of the session is super clear and helpful. But we need to cap the time we spend on this right now.

Please fix the comments that I added and also make this update to the last version of the documentation so that users can read this before 3.0 is released.

Then I'll approve so we can merge this and iterate on it.

SandPod (approved):
🚀

PR #384, opened 2025-12-17
docs: Fix flutter side name confusion

xM1haix:
Keeping the same name across all docs will lead to less confusion.


SandPod (commented):
This is a good clarification, could you make this change to all 3.x.x versions of the documentation along with the non versioned docs?

The non-versioned docs is required to preserve this change when we create new versions of the documentation.

Swiftaxe:
Thank you, @xM1haix. We decided to include the fix for the auth import. The renamings we will leave out for now.

The import fix has been applied across all affected versions (3.0.0–3.4.0 and current docs) in #605, with you credited as co-author.

Thank you for the effort you put into this, and for your patience waiting for a response. We hope to see you contribute again!

PR #438, opened 2026-03-04
docs: Add docs for broken future calls configurations

Crazelu:
Adds documentation for the new configurations introduced to manage broken future calls


marcelomendoncasoares (changes requested):
A few suggestions for improvement to make the conditions of when it occurs and the check/delete behavior clearer for users.

marcelomendoncasoares (approved):
Made a small change on the style to make it more of a prose (less descriptive on the configs).

PR #447, opened 2026-03-24
docs: Added documentation for new reactive future call functionality.

FXschwartz:
Here is the documentation covering new reactive future call feature https://github.com/serverpod/serverpod/pull/4882

### DIFF
--- a/docs/06-concepts/14-scheduling/06-reactive-future-calls.md
+++ b/docs/06-concepts/14-scheduling/06-reactive-future-calls.md
@@ -1,8 +1,8 @@
 # Reactive Future Calls
 
-Reactive future calls let you react to database changes in real-time. When a row is inserted, updated, or deleted, Serverpod can automatically invoke your code with the changed data. This is useful for scenarios like sending notifications when an order is confirmed, syncing data to external systems, or triggering workflows based on state changes.
+Reactive future calls let you react to database changes in near real-time. When a row is inserted, updated, or deleted, Serverpod can automatically and asynchronously invoke your code with the changed data. This is useful for scenarios like sending notifications when an order is confirmed, syncing data to external systems, or triggering workflows based on state changes.
 
-Under the hood, reactive future calls use PostgreSQL triggers and an outbox pattern. A database trigger fires within the same transaction as the data change, writing an entry to an outbox table. Serverpod polls the outbox and dispatches matching entries to your handler. Because the trigger runs in the same transaction, rolled-back changes never produce events.
+Under the hood, reactive future calls use PostgreSQL triggers and an outbox pattern. A database trigger fires within the same transaction as the data change, writing an entry to an outbox table. Serverpod periodically polls the outbox on a configurable scan interval and dispatches matching entries to your handler. Because the trigger runs in the same transaction, rolled-back changes never produce events.
 
 ## Creating a reactive future call
 
@@ -97,7 +97,7 @@ When data changes in a watched table:
 5. Processed entries are deleted from the outbox.
 
 :::info
-The outbox scanner uses the same scan interval as regular future calls, configured via `FutureCallConfig.scanInterval`.
+The outbox scanner uses the same scan interval as regular future calls, configured via the `futureCall.scanInterval` setting in your YAML configuration.
 :::
 
 ## Transaction safety
--- a/docs/06-concepts/14-scheduling/06-reactive-future-calls.md
+++ b/docs/06-concepts/14-scheduling/06-reactive-future-calls.md
@@ -1,0 +1,131 @@
+# Reactive Future Calls
+
+Reactive future calls let you react to database changes in real-time. When a row is inserted, updated, or deleted, Serverpod can automatically invoke your code with the changed data. This is useful for scenarios like sending notifications when an order is confirmed, syncing data to external systems, or triggering workflows based on state changes.
+
+Under the hood, reactive future calls use PostgreSQL triggers and an outbox pattern. A database trigger fires within the same transaction as the data change, writing an entry to an outbox table. Serverpod polls the outbox and dispatches matching entries to your handler. Because the trigger runs in the same transaction, rolled-back changes never produce events.
+
+## Creating a reactive future call
+
+Every model with a database table gets a generated intermediate class that you extend. For example, if you have a `Trip` model, Serverpod generates `TripReactiveFutureCall`:
+
+```dart
+import 'package:serverpod/serverpod.dart';
+import 'src/generated/protocol.dart';
+
+class NotifyPassengersAboutTripConfirmation extends TripReactiveFutureCall {
+  @override
+  WhereExpressionBuilder<TripTable> get where =>
+      (t) => t.status.equals('Confirmed');
+
+  @override
+  Future<void> react(Session session, List<Trip> objects) async {
+    for (final trip in objects) {
+      // Send notifications to passengers
+    }
+  }
+}
+```
+
+The `where` getter defines a condition that becomes a `WHEN` clause on the PostgreSQL trigger. Only changes matching this condition will create outbox entries. In the example above, the trigger only fires when the `status` column equals `'Confirmed'`.
+
+The `react` method is called with all matching rows from a single outbox scan. Rows are batched together for efficiency.
+
+## Detecting column changes
+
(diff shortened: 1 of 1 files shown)


docs/06-concepts/14-scheduling/06-reactive-future-calls.md
```diff
@@ -0,0 +1,131 @@
+# Reactive Future Calls
```
Swiftaxe, line 1: We have started adding frontmatter description to all pages for SEO.
Also, use sentence case for headings.

```suggestion
---
description: React to database row changes in near real-time using Serverpod's reactive future calls, which trigger your handler when inserts, updates, or deletes match a condition.
---

# Reactive future calls
```

docs/06-concepts/14-scheduling/06-reactive-future-calls.md
```diff
@@ -0,0 +1,131 @@
...
+  }
+}
+```
+
+## Code generation
+
+After creating your reactive future call class, run code generation:
+
+```bash
+$ serverpod generate
```
Swiftaxe, line 78: We will update this to match the new serverpod start workflow in a different PR.

docs/06-concepts/14-scheduling/06-reactive-future-calls.md
```diff
@@ -0,0 +1,131 @@
...
+```dart
+// This will NOT trigger react — the transaction is rolled back
+await session.db.transaction((transaction) async {
+  await Trip.db.insertRow(session, Trip(status: 'Confirmed'));
+  throw Exception('Something went wrong');
+});
+
+// This WILL trigger react — the transaction is committed
+await session.db.transaction((transaction) async {
+  await Trip.db.insertRow(session, Trip(status: 'Confirmed'));
```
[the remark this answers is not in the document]
Swiftaxe: When fact-checking this, I find the suggestion by Copilot to be correct.

Wrapping operations in session.db.transaction((transaction) async { ... }) doesn't automatically enlist them in the transaction. 

You have to explicitly opt each database call into the transaction by passing transaction: transaction. Without it, the call runs on its own separate connection and commits immediately. The outer transaction has no hold over it.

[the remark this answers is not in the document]
Swiftaxe: Same here, we should apply it.

docs/06-concepts/14-scheduling/06-reactive-future-calls.md
```diff
@@ -0,0 +1,131 @@
+# Reactive Future Calls
+
+Reactive future calls let you react to database changes in near real-time. When a row is inserted, updated, or deleted, Serverpod can automatically and asynchronously invoke your code with the changed data. This is useful for scenarios like sending notifications when an order is confirmed, syncing data to external systems, or triggering workflows based on state changes.
+
+Under the hood, reactive future calls use PostgreSQL triggers and an outbox pattern. A database trigger fires within the same transaction as the data change, writing an entry to an outbox table. Serverpod periodically polls the outbox on a configurable scan interval and dispatches matching entries to your handler. Because the trigger runs in the same transaction, rolled-back changes never produce events.
```
Swiftaxe, line 5: The users don't need to know how things work under the hood. What they need to know here is that the rolled-back changes never produce events. We can skip directly to the point. 

```suggestion
Because the trigger runs in the same transaction, rolled-back changes never produce events.
```

docs/06-concepts/14-scheduling/06-reactive-future-calls.md
```diff
@@ -0,0 +1,131 @@
...
+
+After creating your reactive future call class, run code generation:
+
+```bash
+$ serverpod generate
+```
+
+Your reactive future calls are automatically discovered and registered alongside regular future calls. No manual registration is needed.
+
+## How it works
```
Swiftaxe, line 83: There are many technical details here that the user probably does not need to concern themselves with.

Consider just saying something like for the whole ## How it works section:

## How it works

Reactive future calls use a scan interval. The same interval as regular future calls, configured via `futureCall.scanInterval`. Serverpod polls for new entries at this interval, so handlers fire in near real-time, rather than instantly.

Rows that accumulate between scans are batched together and delivered to `react` in a single call. This is why `react` receives a `List` rather than a single object.

docs/06-concepts/14-scheduling/06-reactive-future-calls.md
```diff
@@ -0,0 +1,131 @@
...
+When data changes in a watched table:
+
+1. The trigger evaluates the `WHEN` condition.
+2. If matched, the trigger inserts a row into the outbox table within the same transaction.
+3. The outbox scanner picks up new entries on its next poll.
+4. Entries are grouped by handler and dispatched to the `react` method.
+5. Processed entries are deleted from the outbox.
+
+:::info
+The outbox scanner uses the same scan interval as regular future calls, configured via the `futureCall.scanInterval` setting in your YAML configuration.
```
developerjamiu, line 100: Suggestion: this names `futureCall.scanInterval` but doesn't link to the configuration page that documents it. A reader who wants to actually set the value has to navigate the sidebar manually.

```suggestion
The outbox scanner uses the same scan interval as regular future calls, configured via the `futureCall.scanInterval` setting in your YAML configuration. See [Configuration](configuration) for the full set of future-call options.
```

docs/06-concepts/14-scheduling/06-reactive-future-calls.md
```diff
@@ -0,0 +1,131 @@
...
+  @override
+  Future<void> react(Session session, List<Trip> objects) async {
+    for (final trip in objects) {
+      // Send notifications to passengers
+    }
+  }
+}
+```
+
+The `where` getter defines a condition that becomes a `WHEN` clause on the PostgreSQL trigger. Only changes matching this condition will create outbox entries. In the example above, the trigger only fires when the `status` column equals `'Confirmed'`.
```
Swiftaxe, line 29: Mentioning 'outbox' is an implementation details of no concern to our users.

Could we boil these two paragraphs down to the essentials? Perhaps something like: 

The `react` method is your handler. It's called with every row that matches the condition. The `where` getter filters which changes reach the handler.

Swiftaxe (commented):
Some comments added. Mainly about leaving out implementation details and skipping to what is essential for the user to know.

docs/06-concepts/14-scheduling/06-reactive-future-calls.md
```diff
@@ -0,0 +1,131 @@
...
+4. Entries are grouped by handler and dispatched to the `react` method.
+5. Processed entries are deleted from the outbox.
+
+:::info
+The outbox scanner uses the same scan interval as regular future calls, configured via the `futureCall.scanInterval` setting in your YAML configuration.
+:::
+
+## Transaction safety
+
+Because the trigger runs inside the same database transaction as the data change, reactive future calls have transactional guarantees:
```
developerjamiu, line 105: Suggestion: this section talks about transaction commit/rollback semantics without linking to the transactions docs. A reader unfamiliar with `session.db.transaction` has to find it themselves.

```suggestion
Because the trigger runs inside the same database transaction as the data change, reactive future calls have transactional guarantees (see [Transactions](../database/transactions) for the underlying API):
```

docs/06-concepts/14-scheduling/06-reactive-future-calls.md
```diff
@@ -0,0 +1,131 @@
...
+// This will NOT trigger react — the transaction is rolled back
+await session.db.transaction((transaction) async {
+  await Trip.db.insertRow(session, Trip(status: 'Confirmed'));
+  throw Exception('Something went wrong');
+});
+
+// This WILL trigger react — the transaction is committed
+await session.db.transaction((transaction) async {
+  await Trip.db.insertRow(session, Trip(status: 'Confirmed'));
+});
```
developerjamiu, line 120: Nit: em-dashes in code comments. The style guide discourages them in running text, and comments read as prose to the reader. Swap for periods or colons.

```suggestion
// This will NOT trigger react. The transaction is rolled back.
await session.db.transaction((transaction) async {
  await Trip.db.insertRow(session, Trip(status: 'Confirmed'));
  throw Exception('Something went wrong');
});

// This WILL trigger react. The transaction is committed.
await session.db.transaction((transaction) async {
  await Trip.db.insertRow(session, Trip(status: 'Confirmed'));
});
```

PR #454, opened 2026-03-31
docs: Remove $ sign in code examples

### LINKED ISSUE #515: Remove the $ symbol in all md to get copy & paste more easy
Example:

Instead of this

```bash
$ cd mypod/mypod_server
$ docker compose up --build --detach
$ dart bin/main.dart --apply-migrations
```

use this

```bash
cd mypod/mypod_server
docker compose up --build --detach
dart bin/main.dart --apply-migrations
```

Then its is much easy to use the copy & paste function.

Swiftaxe:
Addresses the dollar signs in code examples and closes https://github.com/serverpod/serverpod_docs/issues/515

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Before contributing to our documentation, please read our [style guide](STYLE_GU
 The project includes a Makefile that provides shortcuts for common tasks. You can see all available commands by running:
 
 ```bash
-$ make help
+make help
 ```
 
 This will display a list of all available commands and their descriptions. The Makefile includes commands for installation, starting the development server, creating versions, formatting markdown files, and cleaning build artifacts.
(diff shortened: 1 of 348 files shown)


docs/index.md
```diff
@@ -54,3 +54,135 @@ Serverpod comes packed with powerful features - batteries included.
...
+Serverpod is tested on Mac, Windows, and Linux. Before you can install Serverpod, you need to have **[Flutter](https://flutter.dev/docs/get-started/install)** installed.
+
+:::info
+Check your Flutter installation by running the following command in your terminal:
+
+```bash
+flutter doctor
+```
+
+:::
```
marcelomendoncasoares: Question: This change is actually unrelated from the current PR, right? Might be better to split it so we don't mix up different scopes.

marcelomendoncasoares (changes requested):
Just one question on change that seems unrelated.

Also, should we remove the `$` on past documentations (the `versioned_docs` folder) as well?

marcelomendoncasoares:
Also one more `$` from https://github.com/serverpod/serverpod_docs/pull/438.

marcelomendoncasoares:
Is this still needed after https://github.com/serverpod/serverpod_docs/pull/480?

Swiftaxe:
> Is this still needed after #480?

You are right! Will close this, and the issue.

PR #456, opened 2026-04-01
docs: Moved backward compatibility to the bottom

Swiftaxe:
Moved backward compatibility to the bottom of the reference section


marcelomendoncasoares (commented):
Should we also rename all other pages to keep the numbering consistent?

If we intend to also move other sections around, we should probably do all at once, since it will force conflicts on all existing PRs.

Swiftaxe:
I plan to provide a suggestion for restructuring the entire docs next week. So let's close this PR and make it part of the big restructure instead.

PR #459, opened 2026-04-03
docs: Update documentation for ordering of batch delete returned results

henycave:
Updates documentation to reflect the newly added support for ordering of batch delete returned results [serverpod/pull/4852](https://github.com/serverpod/serverpod/pull/4852)


marcelomendoncasoares (approved):
Thanks for the quick update to the docs, @henycave! 🚀

PR #476, opened 2026-05-01
docs: Adjust menu so working with endpoints and models show as subpages

Swiftaxe:
### Sidebar menu adjustment 

- "Working with endpoints" now shows as sub-items under 'Endpoints' category node
- "Working with models" now shows as sub-items under 'Models' category node

This is consistent with how other sub-items are displayed.

image

### No broken inbound links

Both pages define a slug in their frontmatter that pins the page to a fixed URL regardless of where the file sits in the folder structure. Because the slugs are unchanged, all existing inbound links continue to resolve correctly.


marcelomendoncasoares (approved):
Nice! This is indeed more on the pattern of the rest of the docs.

PR #481, opened 2026-05-06
docs: Fixes to Installation and Quickstart.

vlidholt:


docs/01-installation.md
```diff
@@ -40,7 +40,7 @@ If you are using Docker Desktop and you get an error, make sure that Docker is r
 
 Serverpod is installed using the Dart package manager. To install Serverpod, run the following command in your terminal:
 
-```bash
+```txt
```
marcelomendoncasoares, line 43: Question: Why text instead of bash?

marcelomendoncasoares (approved):
Just one question, but leaving approved.

docs/01-installation.md
vlidholt: The text looked messed up with bash formatting.

PR #485, opened 2026-05-07
docs: Corrections to Middleware page

developerjamiu:
## Summary

Fact-checked the Middleware page. One correction:

- The `Handler` typedef was documented as `Future<Response> Function(Request)` but Relic actually defines it as `FutureOr<Result> Function(Request)`. `Result` is a sealed type with `Response`, `Hijack`, and `WebSocketUpgrade` subclasses, which is why the example's `if (result is Response)` narrowing is required.

Added a one-line note explaining `Result` so readers can see why the type check exists.

## Test plan

- [ ] Render the page locally and confirm the typedef block reads correctly.


Swiftaxe (approved):
LGTM!

PR #486, opened 2026-05-07
docs: Corrections to Working with models page

developerjamiu:
## Summary

Fact-checked the Working with models page. Three corrections:

- **Outdated enum default**: The page framed `byIndex` as the default and recommended setting `byName` because the default would change "in version 3 of Serverpod." That change has already happened — `byName` is the default in 3.5.0-beta.4. Reworded the section so it states the current default and warns about `byIndex` instead.
- **`distanceFunction` keyword row**: Listed only `l2, innerProduct, cosine, l1`, missing `jaccard` and `hamming` (used for `Bit` vectors). Replaced with a pointer to the indexing page, which already documents the full type×index×function matrix.
- **Header style**: `## Default Values` → `## Default values` per [STYLE_GUIDE.md](STYLE_GUIDE.md#L54-L56) sentence-case rule. Anchor stays the same.

## Test plan

- [ ] Render the page locally and confirm the Enum section reads correctly and the keyword table row links work.

### DIFF
--- a/docs/06-concepts/02-models/01-models.md
+++ b/docs/06-concepts/02-models/01-models.md
@@ -169,22 +169,22 @@ values:
   - bird
 ```
 
-By default the serialization will convert the enum to an int representing the index of the value. Changing the order may therefore have unforeseen consequences when reusing old data (such as from a database). Changing the serialization to be based on the name instead of index is easy.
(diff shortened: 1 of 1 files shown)


Swiftaxe (approved):
LGTM!

PR #487, opened 2026-05-07
docs: Corrections to Inheritance and polymorphism page

developerjamiu:
## Summary

Fact-checked the Inheritance and polymorphism page. Three corrections:

- **Generated code example**: Showed `ChildClass extends ParentClass` redeclaring the parent's `String name` field. The analyzer explicitly forbids field redeclaration, and the actual generated code only declares new fields and forwards inherited ones via `super`. Removed the spurious line.
- **`### Extending a Class` → `### Extending a class`**: sentence case per STYLE_GUIDE.md.
- **`### Sealed Classes` → `### Sealed classes`**: same.

## Test plan

- [ ] Render the page locally and confirm the generated code block under "Sealed classes" shows only `int age` on the child.


Swiftaxe (approved):
LGTM!

PR #491, opened 2026-05-11
docs: Corrections to Custom serialization page

developerjamiu:
## Summary

- Fixed `toJson` example: bare `name:` → `'name':` (wouldn't compile as-is).
- `copyWith()` moved out of required rules. It's only needed when the class is used as a field inside a generated `.spy.yaml` model (verified against test fixtures). Added dedicated section.
- Recast "three simple rules" — the list had four items.
- Added missing return type to second `toJson` example.
- `### Implementation Example` → `### Implementation example`.

## Test plan

- [ ] Render locally and confirm the new section reads correctly.


Swiftaxe (approved):
LGTM!

PR #527, opened 2026-05-19
feat: Document the global cache fallback during development

marcelomendoncasoares:
Documentation for https://github.com/serverpod/serverpod/pull/5140.

### DIFF
--- a/docs/06-concepts/08-caching.md
+++ b/docs/06-concepts/08-caching.md
@@ -1,6 +1,6 @@
 # Caching
 
-Accessing the database can be expensive for complex queries or if you need to run many different queries for a specific task. Serverpod makes it easy to cache frequently requested objects in the memory of your server. Any object can be cached, including primitive types (int, String, DateTime, Duration, ByteData, UuidValue), lists, maps, and serializable models. Objects can be stored in the Redis cache if your Serverpod is hosted across multiple servers in a cluster.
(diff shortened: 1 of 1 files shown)


docs/06-concepts/08-caching.md
```diff
@@ -30,10 +30,14 @@ Future<UserData> getUserData(Session session, int userId) async {
...
-In total, there are three caches where you can store your objects. Two caches are local to the server handling the current session, and one is distributed across the server cluster through Redis. There are two variants for the local cache, one regular cache, and a priority cache. Place objects that are frequently accessed in the priority cache.
+In total, there are three caches where you can store your objects. Two caches are local to the server handling the current session, and one is distributed across the server cluster through Redis. There are two variants for the local cache: one regular cache, and a priority cache. Place objects that are frequently accessed in the priority cache.
 
 Depending on the type and number of objects that are cached in the global cache, you may want to specify custom caching rules in Redis. This is currently not handled automatically by Serverpod.
 
+:::info
+During development, you can run and test uses of the global cache on your code without depending on a running Redis server. If the server runs on `development`/`test` run mode and the Redis cache is configured, but not available, the global cache will fall back to an isolated in-memory cache.
```
developerjamiu: Three accuracy points and small wording:
- Fallback fires in any non-production mode, not just `development`/`test` (`lib/src/server/serverpod.dart. around 807`).
- Also fires when Redis isn't configured, not only when configured-but-unavailable (`caches.dart around 25-28`).
- Wording: `runs on` → `runs in`, drop the comma in "configured, but not available", `uses of the global cache on your code` → `code that uses the global cache`.

```suggestion
During development, you can run code that uses the global cache without a running Redis server. If Redis is unreachable (or not configured) in a non-production run mode, the global cache falls back to an isolated in-memory cache.
```

developerjamiu (approved):

PR #532, opened 2026-05-21
docs: Moved all starter pages into the Get started folder

Swiftaxe:
Moved all the starter pages into the "get started" folder to create a single entry point for new users.

Sidebar labels and the first title of each page are also changed to use the active voice.

image

A slug is used to keep the same URL for moved pages, avoiding broken external links.

The source for the number icons could not be found, so all numbers were replaced with a similar-looking font.

Note: updating page content to align with the next release is not part of this PR. It will be pushed separately.


docs/04-get-started/02-quickstart.md
```diff
@@ -1,8 +1,11 @@
 ---
 sidebar_label: Quickstart
-sidebar_class_name: sidebar-icon-quickstart
+sidebar_class_name: sidebar-icon-get-started-step-2
+slug: /quickstart
 ---
 
+<!-- markdownlint-disable MD041 -->
```
developerjamiu, line 7: Why do we have this?

developerjamiu (approved):
This looks good.

docs/04-get-started/02-quickstart.md
Swiftaxe: The markdown lint complains about the page not starting with the title, due to the imports.

PR #539, opened 2026-05-26
docs: Restructure Deploy your app section with Custom hosting

developerjamiu:
## Summary

- "Deploy your app" is now a dropdown with two children: **Deploy to Serverpod Cloud** (recommended path, links out to the Cloud docs) and **Custom hosting** (self-hosted guides).
- Both categories are dropdown-only (no landing page), matching the other sections.
- The self-hosted guides move under `custom-hosting/`, with "Choosing a strategy" first. Inbound links updated.
- Icons: cloud-sun on Deploy to Serverpod Cloud, hammer on Custom hosting.

## Output

image

## Test plan

- [ ] Sidebar: Deploy your app → Deploy to Serverpod Cloud + Custom hosting, both dropdowns.
- [ ] `npm run build` passes (no broken links).

## Issue
Closes: https://github.com/serverpod/serverpod_docs/issues/541

PR #536, opened 2026-05-24
docs: Arrange order of top-level menu items

Swiftaxe:
This arranges the top-level menu items in accordance with the new information architecture.

image

Note: Serverpod Academy page gets added in #500.

Using the sidebar position for the tutorials, rather than renaming folders, prevents renaming the concepts folder, which would cause 130+ additional file changes.

PR #556, opened 2026-05-27
docs: Remove the stale duplicate Deploy your app section

developerjamiu:
#539 merged without rebasing onto #536, leaving two Deploy sections on main: the reworked content under `08-deployments/` (missing its category label) and the old, stale pages still under `09-deployments/` (which held the "Deploy your app" `_category_.json`).

This keeps the reworked section in place and drops the stale duplicate:

- Moves the `"Deploy your app"` `_category_.json` onto the reworked content in `08-deployments/`.
- Removes the old `09-deployments/` pages (already rewritten into `custom-hosting/`).

The reworked content from #539 is unchanged, the inbound link from the Build your first app deploy page stays valid, and `npm run build` passes with no broken links.


Swiftaxe (approved):
LGTM!

PR #560, opened 2026-05-29
docs: Add getting started guide for email authentication setup

Zfinix:
## Summary

New projects on Serverpod 3.5+ already ship with the auth module installed, but the existing `01-setup.md` page is dense and oriented at users adding the module manually. That's a rough first-touch experience for someone who just wants to turn on email sign-in.

This PR:

- Adds a new `01-get-started.md` page that walks a new-project user through the smooth path: turn on the sign-in screen, run `serverpod start`, sign up using the dev console code, sign in.
- Reframes `01-setup.md` as a migration guide for upgrading existing projects, with a note redirecting new-project users to the new page.
- Adds `sidebar_label` on both pages so they don't overflow the sidebar.

A small follow-up PR will rename `01-setup.md` to `02-migrate-auth-module.md` and renumber the rest of the auth section. Kept separate to keep this PR scoped to content.

### DIFF
--- a/docs/06-concepts/11-authentication/01-get-started.md
+++ b/docs/06-concepts/11-authentication/01-get-started.md
@@ -13,8 +13,8 @@ This guide walks you through that, then shows how to test signing up and signing
 ## Prerequisites
 
 - A project created with `serverpod create` on Serverpod 3.5 or later. If you are upgrading an older project, follow [Migrate to the new auth module](./setup) first to add authentication.
-- Docker installed and running, so the server can start its database.
 - The Flutter SDK installed, so you can run the app.
+- Docker installed and running, if your project uses a Docker Postgres. Projects on the embedded Postgres option don't need Docker.
 
 ## Show the sign-in screen
 
--- a/docs/06-concepts/11-authentication/01-get-started.md
+++ b/docs/06-concepts/11-authentication/01-get-started.md
@@ -12,7 +12,7 @@ This guide walks you through that, then shows how to test signing up and signing
 
 ## Prerequisites
 
-- A project created with `serverpod create` on Serverpod 3.5 or later. If you are upgrading an older project, follow [Migrate to the new auth module](./setup) first to add authentication.
+- A project created with `serverpod create` on Serverpod 3.5 or later. For older projects, see [Setup](./setup) first to add the auth module.
 - The Flutter SDK installed, so you can run the app.
 - Docker installed and running, if your project uses a Docker Postgres. Projects on the embedded Postgres option don't need Docker.
 
--- a/docs/06-concepts/11-authentication/01-get-started.md
+++ b/docs/06-concepts/11-authentication/01-get-started.md
@@ -12,7 +12,7 @@ This guide walks you through that, then shows how to test signing up and signing
 
 ## Prerequisites
 
-- A project created with `serverpod create` on Serverpod 3.4 or later. If you are upgrading an older project, follow [Migrate to the new auth module](./setup) first to add authentication.
+- A project created with `serverpod create` on Serverpod 3.5 or later. If you are upgrading an older project, follow [Migrate to the new auth module](./setup) first to add authentication.
 - Docker installed and running, so the server can start its database.
 - The Flutter SDK installed, so you can run the app.
 
--- a/docs/06-concepts/11-authentication/01-get-started.md
+++ b/docs/06-concepts/11-authentication/01-get-started.md
@@ -1,0 +1,73 @@
+---
+sidebar_label: Get started
+---
+
+# Get started with authentication
+
+When you create a project with `serverpod create`, email sign-in is already built in. The server is configured, the database is ready, and your app is connected to it. You only need to turn on the sign-in screen and run the app.
+
+This guide walks you through that, then shows how to test signing up and signing in.
+
+![Sign-in with Serverpod](/img/authentication/sign-in-widget-device.png)
+
+## Prerequisites
+
+- A project created with `serverpod create` on Serverpod 3.4 or later. If you are upgrading an older project, follow [Migrate to the new auth module](./setup) first to add authentication.
+- Docker installed and running, so the server can start its database.
+- The Flutter SDK installed, so you can run the app.
+
+## Show the sign-in screen
+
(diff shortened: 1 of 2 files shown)


docs/06-concepts/11-authentication/01-getting-started.md
```diff
@@ -0,0 +1,92 @@
+# Setup
```
Swiftaxe: Consider a more descriptive title. Setup was also used the old page, which keeps it as a filename, which is a bit conflicting.

docs/06-concepts/11-authentication/01-getting-started.md
```diff
@@ -0,0 +1,92 @@
...
+
+![Sign-in with Serverpod](/img/authentication/sign-in-widget-device.png)
+
+## Prerequisites
+
+- A project created with `serverpod create` on Serverpod 3.4 or later. If you are upgrading an older project, follow [Migrate to the new auth module](./setup) first to add authentication.
+- Docker installed and running, so the server can start its database.
+- The Flutter SDK installed, so you can run the app.
+
+## What you get out of the box
```
Swiftaxe: The whole "What you get out of the box" is a filler not needed. It repeats what is already said at the top. Remove it.

docs/06-concepts/11-authentication/01-getting-started.md
```diff
@@ -0,0 +1,92 @@
...
+
+That is the only change your app needs.
+
+## Start the database
+
+From the server directory, start the database and apply the migration that creates the sign-in tables:
+
+```bash
+cd my_project_server
+docker compose up --build --detach
```
Swiftaxe: Docker and dart run is not relevant for 3.5. Update to conform to latest version.

docs/06-concepts/11-authentication/01-getting-started.md
```diff
@@ -0,0 +1,92 @@
...
+
+This also starts the server.
+
+## Run the app
+
+With the server running, start your app:
+
+```bash
+cd my_project_flutter
+flutter run
```
Swiftaxe: No Flutter run needed in latest version.

Swiftaxe (commented):
Looks clean! Added a few comments.

docs/06-concepts/11-authentication/01-setup.md
```diff
@@ -1,6 +1,10 @@
-# Setup
+# Migrate to the new auth module
```
Swiftaxe: If this will be the migration guide, we need to fully test it as that and take the other attempts at migration guides into consideration (ie the old PR still hanging)

docs/06-concepts/11-authentication/01-getting-started.md
Zfinix: Removed

docs/06-concepts/11-authentication/01-get-started.md
```diff
@@ -0,0 +1,73 @@
...
+
+When you create a project with `serverpod create`, email sign-in is already built in. The server is configured, the database is ready, and your app is connected to it. You only need to turn on the sign-in screen and run the app.
+
+This guide walks you through that, then shows how to test signing up and signing in.
+
+![Sign-in with Serverpod](/img/authentication/sign-in-widget-device.png)
+
+## Prerequisites
+
+- A project created with `serverpod create` on Serverpod 3.4 or later. If you are upgrading an older project, follow [Migrate to the new auth module](./setup) first to add authentication.
```
developerjamiu: Suggestion: the pre-req says "Serverpod 3.4 or later" but the workflow below uses `serverpod start`, which only exists from 3.5. On 3.4, readers would need to start Docker, the server, and the Flutter app separately. Recommend bumping the pre-req to 3.5. Same fix in `01-setup.md` line 9.

```suggestion
- A project created with `serverpod create` on Serverpod 3.5 or later. If you are upgrading an older project, follow [Migrate to the new auth module](./setup) first to add authentication.
```

docs/06-concepts/11-authentication/01-get-started.md
```diff
@@ -0,0 +1,73 @@
...
+- Docker installed and running, so the server can start its database.
+- The Flutter SDK installed, so you can run the app.
```
developerjamiu: Suggestion: Serverpod 3.5 ships with an embedded Postgres option, so Docker isn't required for every project. Recommend softening this so readers on the embedded path don't think they need Docker.

```suggestion
- The Flutter SDK installed, so you can run the app.
- Docker installed and running, if your project uses a Docker Postgres. Projects on the embedded Postgres option don't need Docker.
```

Order swapped because Flutter is the universal pre-req and Docker is conditional.

developerjamiu (changes requested):
The writeup is solid and flows well; the new Get started page reads cleanly and hits the goal of a simple email-auth path. Nice work.

Reinforcing @Swiftaxe earlier note on the rename of `01-setup.md` to "Migrate to the new auth module": the title now promises content the body doesn't have. The steps still teach a fresh install rather than a migration from serverpod_auth, so a reader coming from an older project will follow them, install serverpod_auth_idp, and still have an old-auth schema with no path forward.

This is not blocking as it was mentioned earlier that there's an existing PR for the actual migration but flagging it so we address this soon.

developerjamiu (approved):

Zfinix:
> The writeup is solid and flows well; the new Get started page reads cleanly and hits the goal of a simple email-auth path. Nice work.
> 
> Reinforcing @Swiftaxe earlier note on the rename of `01-setup.md` to "Migrate to the new auth module": the title now promises content the body doesn't have. The steps still teach a fresh install rather than a migration from serverpod_auth, so a reader coming from an older project will follow them, install serverpod_auth_idp, and still have an old-auth schema with no path forward.
> 
> This is not blocking as it was mentioned earlier that there's an existing PR for the actual migration but flagging it so we address this soon.

Yes I agree, I have dropped the edits to 01-setup.md from this PR, the file is back to its main state. will tackle migration in a separate PR

developerjamiu (approved):

Swiftaxe (approved):
Great entry-path into authentication!

LGTM!

Issue #565, opened 2026-06-02
Address all mentions of dart run bin/main.dart


Swiftaxe:
It is still possible to run bin/main.dart, so it should be documented, but it's no longer the main workflow.

The docs should reflect the new workflow. 

Perhaps possible to use AI to update these in batches, replacing it with pressing 'R' for hot-reload.

PR #599, opened 2026-06-12
docs: Clarify how migrations are applied with serverpod start

Zfinix:
## Summary

Notes that `serverpod start` applies migrations during local development (on first boot, or by pressing **A** in the terminal UI). The explicit `--apply-migrations` runtime command stays documented for production and automated pipelines.

The remaining `dart run bin/main.dart` mentions are intentional: the `--role maintenance` and production invocations are runtime calls that `serverpod start` should not replace (it is a development tool and always applies migrations).

Part of #565

### DIFF
--- a/docs/06-concepts/06-database/11-migrations.md
+++ b/docs/06-concepts/06-database/11-migrations.md
@@ -94,7 +94,9 @@ For each migration, five files are created:
 
 ## Apply migrations
 
-Migrations are applied using the server runtime. To apply migrations, navigate to your project's `server` package directory, then start the server with the `--apply-migrations` flag. Migrations are applied as part of the startup sequence and the framework asserts that each migration is only applied once to the database.
+During local development, `serverpod start` applies your migrations for you: pending migrations run on the first boot, and you can apply new ones at any time by pressing **A** in the terminal UI.
+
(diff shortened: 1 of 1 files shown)


docs/06-concepts/06-database/11-migrations.md
```diff
@@ -94,7 +94,9 @@ For each migration, five files are created:
 
 ## Apply migrations
 
-Migrations are applied using the server runtime. To apply migrations, navigate to your project's `server` package directory, then start the server with the `--apply-migrations` flag. Migrations are applied as part of the startup sequence and the framework asserts that each migration is only applied once to the database.
+During local development, `serverpod start` applies your migrations for you: pending migrations run on the first boot, and you can apply new ones at any time by pressing **A** in the terminal UI.
```
Swiftaxe: Nit: Users scan documentation real quick. When scanning quickly, the text "pressing A" disappears and what stands out is the "dart run bin/main.dart...". Consider following the pattern in the "Build your first app", where Press **R** stands out by using a bullet.

Image

The same goes for "production and automated pipeline" which disappears in text. Consider if this should be a heading, to catch the attention of users searching for those particular use cases.

Zfinix: Done, split it into During development (bullets so A stands out) and In production or CI headings.

developerjamiu (approved):
LGTM

Swiftaxe (approved):
Nice and small PR!

LGTM!

PR #16, opened 2023-07-19
Alexs readthrough comments

SandPod:

PR #617, opened 2026-06-17
fix: Fact-check of exceptions and sessions

Swiftaxe:
## Summary

- `04-exceptions`: added missing `description` frontmatter, fixed missing `await` on endpoint call in catch example (bug — without it the exception is never caught), removed filler and redundant sentences, removed example narration bullets
- `05-sessions`: added missing `description` frontmatter, replaced internal-mechanics intro with one actionable sentence, fixed all numbered link prefixes (noise pattern #16), trimmed lifecycle and logging sections to remove internal mechanics


Zfinix (approved):

Zfinix:
LGTM

PR #624, opened 2026-06-17
fix: Fact-check of scheduling

Swiftaxe:
## Summary

Fact-checked, removed noise, and added frontmatter to 5 scheduling pages.

**All 5 files:**
- Added missing `description` frontmatter

`01-setup`:
- Removed stray `import 'package:serverpod_auth_idp_server/core.dart';` from the `server.dart` example (unrelated auth import)

`02-recurring-task`:
- Fixed "continously" → "continuously"
- Removed same stray auth import and stripped trailing spaces from the code block

`03-inheritance`:
- "augment the superclass's one" → "overriding the superclass implementation"

`04-configuration`:
- Removed redundant "Below is an example of how you can configure future calls in a YAML file:" lead-in
- Removed 5× repeated "Example configuration:" labels
- Fixed `checkBrokenCalls` and `deleteBrokenCalls` examples — they were shown at the top level but should be nested under `futureCall:` per the configuration reference (factual fix)

`05-legacy`:
- Removed "is simple" filler
- Trimmed verbose registration paragraph
- "You are now able to register a future call to be invoked in the future by calling either..." → "With the future call registered, you can schedule it using either..."
- `DateTime(2025, 1, 1)` → `DateTime(2030, 1, 1)` (past date)
- Fixed note text (lowercase start, unclear reference to "model folder")

### DIFF
--- a/docs/06-concepts/14-scheduling/01-setup.md
+++ b/docs/06-concepts/14-scheduling/01-setup.md
@@ -1,5 +1,5 @@
 ---
-description: Schedule future work in Serverpod using future calls — invoke methods after a delay, at a specific time, or on a recurring interval, with persistence across restarts.
+description: Schedule future work in Serverpod using future calls: invoke methods after a delay, at a specific time, or on a recurring interval, with persistence across restarts.
 ---
 
 # Setup
--- a/docs/06-concepts/14-scheduling/01-setup.md
+++ b/docs/06-concepts/14-scheduling/01-setup.md
@@ -1,3 +1,7 @@
+---
+description: Schedule future work in Serverpod using future calls — invoke methods after a delay, at a specific time, or on a recurring interval, with persistence across restarts.
+---
+
 # Setup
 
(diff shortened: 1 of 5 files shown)


docs/06-concepts/14-scheduling/01-setup.md
```diff
@@ -1,3 +1,7 @@
+---
+description: Schedule future work in Serverpod using future calls — invoke methods after a delay, at a specific time, or on a recurring interval, with persistence across restarts.
```
developerjamiu: Suggestion: em dash slipped in here. A colon keeps the same structure.

```suggestion
description: Schedule future work in Serverpod using future calls: invoke methods after a delay, at a specific time, or on a recurring interval, with persistence across restarts.
```

docs/06-concepts/14-scheduling/05-legacy.md
```diff
@@ -1,10 +1,14 @@
+---
+description: The legacy string-based API for registering and scheduling future calls, prefer the type-safe API for new code.
```
developerjamiu: Suggestion: the comma joins two clauses where a period would read cleaner.

```suggestion
description: The legacy string-based API for registering and scheduling future calls. Prefer the type-safe API for new code.
```

developerjamiu (approved):

PR #625, opened 2026-06-17
fix: Fact-check of streams, server events, and backward compatibility

Swiftaxe:
## Summary

Fact-checked, removed noise, and added frontmatter to 3 pages.

**All 3 files:**
- Added missing `description` frontmatter

`15-streams`:
- Removed "Note that" weak lead-in from tip
- "utilizing the authentication data" → "using"
- Trimmed verbose auth revocation sentence
- "It's often handy to associate a state" → direct imperative sentence
- Fixed prose naming `connectWebSocket` when the code uses `openStreamingConnection` (factual fix)
- Removed trailing blank line inside code block

`16-server-events`:
- "the message published on the channel" → "the message **is** published" (grammar fix)
- Removed "Receiving messages is just as simple as sending them!" (informal + exclamation mark)
- "processes incoming requests" → "processes incoming messages"
- "you simply cancel" → "cancel"
- Removed "In the above example, the stream is first created and then canceled." (redundant narration)
- "regardless if" → "regardless of whether"

`17-backward-compatibility`:
- "Additionally all the other" → "Additionally, all the other" (missing comma)

### DIFF
--- a/docs/06-concepts/15-streams.md
+++ b/docs/06-concepts/15-streams.md
@@ -1,3 +1,7 @@
+---
+description: Stream data between server and client in Serverpod using streaming methods over a self-managed WebSocket connection, with support for bidirectional streams and serializable exceptions.
+---
+
 # Streams
 
 For some applications, it's not enough to be able to call server-side methods. You may also want to push data from the server to the client or send data two-way. Examples include real-time games or chat applications. Luckily, Serverpod supports a framework for streaming data. It's possible to stream any serialized objects to or from any endpoint.
@@ -53,7 +57,7 @@ In the example above, the `echoStream` method passes back any message sent throu
 
 :::tip
 
-Note that we can mix different types in the stream. This stream is defined as dynamic and can contain any type that can be serialized by Serverpod.
+The stream is defined as dynamic and can contain any type that can be serialized by Serverpod.
 
 :::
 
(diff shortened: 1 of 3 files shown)


docs/06-concepts/15-streams.md
```diff
@@ -75,7 +79,7 @@ All streams in parameters are closed when the method call is over.
 
 Authentication is seamlessly integrated into streaming method calls. When a client initiates a streaming method, the server automatically authenticates the session.
```
developerjamiu: Suggestion: Drop "seamlessly" (flagged marketing vocab). The next sentence already states the real behavior, so the word adds nothing. Optional.

```suggestion
Authentication is integrated into streaming method calls. When a client initiates a streaming method, the server automatically authenticates the session.
```

docs/06-concepts/15-streams.md
```diff
@@ -1,3 +1,7 @@
+---
+description: Stream data between server and client in Serverpod using streaming methods over a self-managed WebSocket connection, with support for bidirectional streams and serializable exceptions.
```
developerjamiu, line 2: Nit: The page body uses "web socket" throughout (e.g. lines 15 and 159), but this description uses "WebSocket". "WebSocket" is the correct brand form, so ideally the body standardizes to it, though that's outside this PR's scope. Flagging so the mismatch is a deliberate choice rather than an oversight.

developerjamiu (approved):
LGTM. Clean fact-check and de-noise pass. Two optional notes.

PR #660, opened 2026-06-24
docs: Document the `unique(per...)` index shorthand

marcelomendoncasoares:
Docs for the feature introduced with PR https://github.com/serverpod/serverpod/pull/5343.


docs/06-concepts/06-database/04-indexing.md
```diff
@@ -59,6 +59,46 @@ fields:
   name: String, unique
 ```
 
+### Composite unique constraints
+
+When a value should be unique only within a scope — for example, a setting key that is unique per user — use `unique(per=...)` on the field that must be unique within that scope. Serverpod auto-generates a composite unique index with the `per` columns first, followed by the annotated field.
```
developerjamiu: Suggestion: we keep em dashes out of the docs. Parentheses read cleanly here:

```suggestion
When a value should be unique only within a scope (for example, a setting key that is unique per user), use `unique(per=...)` on the field that must be unique within that scope. Serverpod auto-generates a composite unique index with the `per` columns first, followed by the annotated field.
```

docs/06-concepts/06-database/04-indexing.md
```diff
@@ -59,6 +59,46 @@ fields:
...
+class: UserSetting
+table: user_setting
+fields:
+  tenantId: int
+  category: String
+  sku: String
+indexes:
+  product_unique_idx:
+    fields: tenantId, category, sku
+    unique: true
```
developerjamiu: This is the expanded form of the `Product` example above, but the class and table are still `UserSetting` (the fields and index name are already `product`). Looks like a copy-paste slip:

```suggestion
class: Product
table: product
fields:
  tenantId: int
  category: String
  sku: String
indexes:
  product_unique_idx:
    fields: tenantId, category, sku
    unique: true
```

docs/06-concepts/06-database/04-indexing.md
marcelomendoncasoares: It is a sad thing that I have always used em-dashes my entire life, which was elegant before it became a signal for cheap AI slop... I'll use parenthesis in the next docs.

docs/06-concepts/02-models/01-models.md
```diff
@@ -547,7 +547,7 @@ fields:
 | [**type (index)**](database/indexing)                               | The type of index to create.                                                                                  |       ✅        |                         |               |
 | [**parameters (index)**](database/indexing#vector-indexes)          | Parameters for specialized index types like HNSW and IVFFLAT vector indexes.                                  |       ✅        |                         |               |
 | [**distanceFunction (index)**](database/indexing#vector-indexes)    | Distance function for vector indexes. Allowed values depend on the field's vector type and index type — see [vector indexes](database/indexing#vector-indexes). |       ✅        |                         |               |
-| [**unique**](database/indexing)                                     | Boolean flag to make the entries unique in the database.                                                      |       ✅        |                         |               |
+| [**unique**](database/indexing)                                     | Creates a unique index. |       ✅        |                         |               |
```
developerjamiu, line 550: Suggestion: since this PR adds the scoped form, mentioning it here helps readers scanning the keyword table find it:

```suggestion
| [**unique**](database/indexing)                                     | Creates a unique index, optionally scoped with `unique(per=...)`. |       ✅        |                         |               |
```

docs/06-concepts/06-database/04-indexing.md
```diff
@@ -59,6 +59,46 @@ fields:
...
+```yaml
+class: Product
+table: product
+fields:
+  tenantId: int
+  category: String
+  sku: String, unique(per=[tenantId, category])
+```
+
+You can also use the equivalent expanded form:
```
developerjamiu: Comment: the shorthand names the index for you, so a reader might not expect the `product_unique_idx` name below. Worth noting the expanded form is where you name it:

```suggestion
You can also use the equivalent expanded form, where you name the index yourself:
```

docs/06-concepts/02-models/01-models.md
marcelomendoncasoares: This one I preferred not to expand because the `unique` keyword assumes three different roles in the models:
- Under an `indexes` entry, it is a boolean flag.
- On a model field without parenthesis, it generates a unique index on that column.
- On a model field with parenthesis and the `per=...` argument, it generates a multi-column unique index.

Adding this detail here seems to ignore the first form under `indexes` entries.

developerjamiu (approved):

PR #661, opened 2026-06-25
docs: Document GiST and SP-GiST geography indexes

charlesarchibong:
Documents the new PostGIS geography support, mainly the GiST and SP-GiST spatial index work from serverpod/serverpod#5392, and fixes a couple of things in the existing geography docs while I was in there.

What's in here:

- The Geography indexes section in the indexing page now covers the `gist` and `spgist` types, `gist` as the default when no type is given, the type restriction and the no-mixing rule, with the exact code-generation error strings quoted. Added an `spgist` example and a note that SP-GiST needs a recent PostGIS.
- Fixed the spatial filter operator name in the filter docs. It said `dwithin`, but the real API is `distanceWithin`. Checked against the serverpod source.
- Added the SRID detail to the geography fields page (fixed to 4326, exposed as `Geography.defaultSrid`, no per-column SRID yet) plus a SQLite limitation note (stored as EWKT text, spatial ops throw at query time).

### DIFF
--- a/docs/06-concepts/06-database/04-indexing.md
+++ b/docs/06-concepts/06-database/04-indexing.md
@@ -305,7 +305,7 @@ indexes:
 A spatial index accelerates all spatial operations (`intersects`, `distanceWithin`, `distance`, `contains`, `within`). For tables with many rows and frequent spatial queries, adding one is strongly recommended.
 :::
 
-Geography fields only support the `gist` and `spgist` index types. Specifying any other type fails code generation with `The "type" property must be one of: gist, spgist.` An index may cover several geography columns, but geography and non-geography fields cannot be mixed in the same index — doing so fails with `Mixing geography and non-geography fields in the same index is not allowed.`
+Geography fields only support the `gist` and `spgist` index types. Specifying any other type fails code generation with `The "type" property must be one of: gist, spgist.` An index may cover several geography columns, but it cannot mix geography and non-geography fields. Doing so fails with `Mixing geography and non-geography fields in the same index is not allowed.`
 
 :::info
 `spgist` indexes on the geography type require a recent version of PostGIS. If your PostgreSQL instance ships an older PostGIS, use `gist` instead.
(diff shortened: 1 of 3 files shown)


docs/06-concepts/06-database/04-indexing.md
```diff
@@ -282,6 +287,26 @@ indexes:
...
+    fields: boundary
+    type: spgist
+```
+
 :::tip
-A GIST index on a geography column accelerates all spatial operations (`intersects`, `dwithin`, `distance`, `contains`, `within`). For tables with many rows and frequent spatial queries, adding a GIST index is strongly recommended.
+A spatial index accelerates all spatial operations (`intersects`, `distanceWithin`, `distance`, `contains`, `within`). For tables with many rows and frequent spatial queries, adding one is strongly recommended.
+:::
+
+Geography fields only support the `gist` and `spgist` index types. Specifying any other type fails code generation with `The "type" property must be one of: gist, spgist.` An index may cover several geography columns, but geography and non-geography fields cannot be mixed in the same index — doing so fails with `Mixing geography and non-geography fields in the same index is not allowed.`
```
developerjamiu: Nice catch on the `dwithin` → `distanceWithin` fix, and the error strings match the source. One small thing: we keep em dashes out of the docs. Splitting this into two sentences fixes it (and reads a touch cleaner):

```suggestion
Geography fields only support the `gist` and `spgist` index types. Specifying any other type fails code generation with `The "type" property must be one of: gist, spgist.` An index may cover several geography columns, but it cannot mix geography and non-geography fields. Doing so fails with `Mixing geography and non-geography fields in the same index is not allowed.`
```

developerjamiu (changes requested):
Good changes. Just the tiny comments and we are good 🚀

docs/06-concepts/06-database/04-indexing.md
charlesarchibong: Good catch, dropped the em dash and split it into two sentences per your suggestion. Pushed in 56d38b9.

charlesarchibong: This is fixed. Thank you

charlesarchibong: @developerjamiu, kindly review.

developerjamiu (approved):
LGTM

PR #663, opened 2026-06-30
docs: Clarify secret access, production storage, and Cloud secrets

### LINKED ISSUE #653: Clarify secrets: startup vs endpoint access, and Serverpod Cloud
## Demand

Part of the Kapa "Secrets & Seeding" cluster — 15 conversations / 15 users.

## Problem

- The docs do not distinguish `pod.getPassword()` (available at server startup) from `session.passwords[...]` (used inside endpoints) — users ask things like "can I use pod.getPassword in Flutter?".
- No guidance on managing secrets on Serverpod Cloud (the managed environment), or where credential files like `google_client_secret.json` / `firebase_service_account_key.json` should live.
- No production storage / gitignore guidance.

## Scope

- `07-configuration.md`: an explicit startup-vs-endpoint note.
- Deployment docs: a "Secrets on Serverpod Cloud" section.
- Production storage / gitignore guidance.

## Verify before writing

- The recommended secret-loading approach on Serverpod Cloud (confirm with the team).

developerjamiu:
Closes #653.

The configuration docs explained how to *declare* secrets but not how to *read* them, which left a few recurring questions unanswered (the most common: "can I use `getPassword` in my Flutter app?"). This adds three subsections to the Secrets section of `07-configuration.md`:

- **Accessing Secrets in Code** — secrets are server-side only and never reach the Flutter app. Read them with `session.passwords['name']` / `session.serverpod.getPassword('name')` inside endpoints, or `pod.getPassword('name')` at startup. The `stripeApiKey` example threads through from where it's declared above.
- **Secrets in Production** — `config/passwords.yaml` and credential files are gitignored by default; in production, set secrets via `SERVERPOD_PASSWORD_*` env vars or your host's secret manager.
- **Secrets on Serverpod Cloud** — short pointer to `scloud password set` (with `--from-file` for JSON credentials), linking to the Cloud passwords/secrets reference for the full story.

Scope is deliberately the configuration page plus a cross-link; the Cloud side is already covered in the Cloud docs, so this doesn't duplicate it.

### Verification
- `getPassword` / `session.passwords` against framework source (`serverpod.dart`, `session.dart`); confirmed no `getPassword` in `serverpod_client` (server-side only).
- Default `.gitignore` entries for `passwords.yaml` and `firebase_service_account_key.json` against the starter template.
- The Cloud `scloud password set` / `--from-file` round-trip run against a live Cloud project (set, listed as `SET (User)`, encrypted).
- Docs build passes; no broken links.

### DIFF
--- a/docs/06-concepts/07-configuration.md
+++ b/docs/06-concepts/07-configuration.md
@@ -83,7 +83,7 @@ These can be separately declared for each run mode in the corresponding yaml fil
 
 ### Secrets
 
-Secrets are declared in the `passwords.yaml` file. The password file is structured with a common `shared` section, any secret put here will be used in all run modes. The other sections are the names of the run modes followed by respective key/value pairs. You can also define custom secrets using [environment variables](#2-via-environment-variables).
+Secrets are declared in the `passwords.yaml` file. Serverpod's API reads them with `getPassword`, so this page uses *secret* and *password* interchangeably. The password file is structured with a common `shared` section, any secret put here will be used in all run modes. The other sections are the names of the run modes followed by respective key/value pairs. You can also define custom secrets using [environment variables](#2-via-environment-variables).
 
 #### Built-in Secrets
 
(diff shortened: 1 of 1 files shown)


docs/06-concepts/07-configuration.md
```diff
@@ -142,18 +142,46 @@ To define a custom password through an environment variable:
...
+```dart
+// `pod` is the Serverpod instance created in run().
+final stripeApiKey = pod.getPassword('stripeApiKey');
+```
+
+This works for built-in and custom secrets alike, whether they come from the passwords file or an environment variable.
+
+#### Secrets in Production
+
+A new project gitignores `config/passwords.yaml` and credential files such as `config/firebase_service_account_key.json`, so secrets are not committed by default. Keep production secrets out of source control.
```
Zfinix: Small nit, reads cleaner without "gitignores" as a verb:
```suggestion
A new project's `.gitignore` excludes `config/passwords.yaml` and credential files such as `config/firebase_service_account_key.json`, so secrets are not committed by default. Keep production secrets out of source control.
```

docs/06-concepts/07-configuration.md
```diff
@@ -142,18 +142,46 @@ To define a custom password through an environment variable:
...
+
+In production, set secrets through `SERVERPOD_PASSWORD_*` environment variables, or your host's secret manager, rather than a checked-in passwords file.
+
+#### Secrets on Serverpod Cloud
```
Zfinix: This uses `scloud password set`, which is Cloud's *Passwords* tier, but the page it links to defines "Secrets" as a separate tier (encrypted values injected as named env vars, not `getPassword`). So readers land on "Secrets" meaning something different from what they just set. The mapping is right, just the heading word collides. Could we retitle to "Passwords on Serverpod Cloud," or note that `getPassword` maps to Cloud's Passwords tier?

Zfinix (approved):

PR #666, opened 2026-07-01
docs: Cover MCP setup in the 4.0 migration guide

### LINKED ISSUE #583: Migration guide does not cover setting up the MCPs
The MCPs powering the agentic workflow are configured when creating a new project. Migrating users need a smooth way to set it up.

The user will need to run `serverpod create .` to install and configure MCP's and install skills.
Also, in Cursor (and maybe other IDEs) you might need to enable the Dart and Serverpod MCP's.

Zfinix:
## Summary

The 3.5 migration guide told upgrading users to set up the agentic workflow with `dart install skills` + `skills get`, which installs the skills but never registers the MCP servers. New projects get their MCP config from `serverpod create`, so migrating users were left without the Serverpod and Dart MCP servers.

This updates the "Set up the agent workflow" section to lead with `serverpod create .`, which configures both MCP servers and installs the skills in one step (the same setup a new project gets). It also:

- Shows the generated `.mcp.json` and notes VS Code's `.vscode/mcp.json` uses a `servers` key instead of `mcpServers`.
- Keeps `dart install skills` / `skills get` as a skills-only refresh path.
- Fixes the quickstart Cursor note to name both the Serverpod and Dart MCP servers (there are two, not one).
- Updates the matching "What's new" bullet and troubleshooting entry.

Verified against the Serverpod CLI 3.5.0-beta.10: `serverpod create .` is non-destructive on an existing project (runs the upgrade flow, configures the MCP servers, installs skills), and the `--ide` editor list and `.mcp.json` structure match the live output.

Closes #583

### DIFF
--- a/docs/04-get-started/02-quickstart.md
+++ b/docs/04-get-started/02-quickstart.md
@@ -64,7 +64,7 @@ Next, open your newly created project with your favorite AI powered editor. Open
 
 :::important
 
-If you are using Cursor, you will need to **enable the MCP server** in your project settings (_Cursor Settings_ > _Tools & MCPs_).
+If you are using Cursor, you will need to **enable the Serverpod and Dart MCP servers** in your project settings (_Cursor Settings_ > _Tools & MCPs_).
 
 :::
 
--- a/docs/11-upgrading/01-upgrade-to-four.md
+++ b/docs/11-upgrading/01-upgrade-to-four.md
@@ -141,15 +141,52 @@ The first run compiles the native build hooks (this can take about 30 seconds) a
 
 Beyond the server, `serverpod start` also launches the project's Flutter apps configured with `auto_launch: true`. For IDE debugging, projects scaffolded with 4.0 include a `launch.json` that runs `serverpod start` with the debugger attached; you can copy that file into your existing project from a fresh 4.0 scaffold if you want the same setup.
 
-## Add the agent skills (optional)
+## Set up the agent workflow (optional)
 
-Version 4.0 ships AI agent skills (for editors like Claude Code and Cursor) and an MCP server. Install them with:
+Version 4.0 ships AI agent skills and MCP servers (for editors like Claude Code and Cursor) that let your agent build, run, and inspect your server. A new project configures these during `serverpod create`, so the smoothest way to add the same setup to your upgraded project is to run the same command against the current directory.
+
+:::warning
+
+Configuration files you created manually can be overwritten, so commit your work before running the command.
+
+:::
+
+From the project's root folder, run:
+
+```bash
+$ serverpod create .
+```
+
+Serverpod detects the existing project and adds the missing pieces without touching your source code. It registers the **Serverpod** and **Dart** MCP servers and installs the agent skills for the editors you select. Each selected editor gets a config file in its own format: `.mcp.json` for Claude, `.cursor/mcp.json` for Cursor, and `.vscode/mcp.json` for VS Code. For Claude, `.mcp.json` looks like this:
+
+```json
+{
+  "mcpServers": {
+    "serverpod": {
+      "command": "serverpod",
+      "args": ["mcp-server", "--server-dir", "<project>_server"]
+    },
+    "dart": {
+      "command": "dart",
+      "args": ["mcp-server"]
+    }
+  }
+}
+```
+
+VS Code's `.vscode/mcp.json` registers the same two servers but nests them under a `servers` key instead of `mcpServers`.
+
+If you are using Cursor, enable the **Serverpod** and **Dart** MCP servers in your project settings (_Cursor Settings_ > _Tools & MCPs_).
+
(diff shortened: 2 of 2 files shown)


docs/11-upgrading/01-upgrade-to-three-five.md
```diff
@@ -141,15 +141,50 @@ The first run compiles the native build hooks (this can take about 30 seconds) a
...
+      "command": "dart",
+      "args": ["mcp-server"]
+    }
+  }
+}
+```
+
+VS Code's `.vscode/mcp.json` registers the same two servers but nests them under a `servers` key instead of `mcpServers`.
+
+:::warning
```
developerjamiu: Suggestion: this warning to commit your work sits below the `serverpod create .` block, so a reader who copies the command first can run it before seeing the caution. It reads better moved just above the command block (around line 148), so the warning comes before the action. The text itself is fine; it is just a block move.

docs/11-upgrading/01-upgrade-to-three-five.md
```diff
@@ -141,15 +141,50 @@ The first run compiles the native build hooks (this can take about 30 seconds) a
...
+## Set up the agent workflow (optional)
 
-Version 3.5 ships AI agent skills (for editors like Claude Code and Cursor) and an MCP server. Install them with:
+Version 3.5 ships AI agent skills and MCP servers (for editors like Claude Code and Cursor) that let your agent build, run, and inspect your server. A new project configures these during `serverpod create`, so the smoothest way to add the same setup to your upgraded project is to run the same command against the current directory. From the project's root folder, run:
+
+```bash
+$ serverpod create .
+```
+
+Serverpod detects the existing project and adds the missing pieces without touching your source code. It registers the **Serverpod** and **Dart** MCP servers and installs the agent skills for the editor you select. The registration is written to `.mcp.json`, along with editor-specific files (`.cursor/mcp.json` and `.vscode/mcp.json`) in each editor's own format. The `.mcp.json` file looks like this:
```
developerjamiu: Suggestion: `.mcp.json` is Claude's own file, not a shared one. In `ide.dart`, each selected editor gets its own file (`.mcp.json` for Claude, `.cursor/mcp.json` for Cursor, `.vscode/mcp.json` for VS Code), so a Cursor-only user gets no `.mcp.json` at all and could go looking for a file that is not there. Reframing avoids that (and fixes "editor" to "editors", since you can select more than one):

```suggestion
Serverpod detects the existing project and adds the missing pieces without touching your source code. It registers the **Serverpod** and **Dart** MCP servers and installs the agent skills for the editors you select. Each selected editor gets a config file in its own format: `.mcp.json` for Claude, `.cursor/mcp.json` for Cursor, and `.vscode/mcp.json` for VS Code. For Claude, `.mcp.json` looks like this:
```

docs/11-upgrading/01-upgrade-to-three-five.md
```diff
@@ -141,15 +141,50 @@ The first run compiles the native build hooks (this can take about 30 seconds) a
 
 Beyond the server, `serverpod start` also launches the project's Flutter app. By default it uses Flutter's `web-server` device and opens your browser. Pass `--flutter-device <name>` to target a specific device. For IDE debugging, projects scaffolded with 3.5 include a `launch.json` that runs `serverpod start` with the debugger attached; you can copy that file into your existing project from a fresh 3.5 scaffold if you want the same setup.
```
developerjamiu: Comment (pre-existing, outside this diff): two stale lines worth a quick fix while this file is open. This line mentions `--flutter-device <name>`, which is not a real `serverpod start` flag (only `--flutter` / `--no-flutter` exist). And line 25 pins `dart install serverpod_cli 3.5.0-beta.9`, while the latest 3.5 beta is `beta.12`. Up to you whether to fold these in here or leave them for a follow-up.

Zfinix: is it not 4.0.0-beta.0 ?

developerjamiu: Oh, yeah it should be 4.0.0-beta.0

docs/11-upgrading/01-upgrade-to-three-five.md
Zfinix: Done, moved it above the command

docs/11-upgrading/01-upgrade-to-three-five.md
Zfinix: Done, applied your reframe

docs/11-upgrading/01-upgrade-to-three-five.md
Zfinix: Rebased onto main, both are fixed there now (4.0.0-beta.0 pin and the flutter line)

developerjamiu (approved):
LGTM

PR #674, opened 2026-07-06
docs: Extract Model reference and reshape Models concept pages

developerjamiu:
Phase 2 IA, PR 3a of 7 (tracked in #671): reshapes the Models subgroup and extracts a Model reference. This is the first half of the Data & the database work; the Database subgroup follows in PR 3b.

## What changed

- **Model reference (new)**: the master model-keyword table moves out of Working with models into a dedicated reference. All 31 keyword rows are preserved, and every link is rebased. This becomes the single source that PR 3b's Database tables page defers to.
- **Working with models**: the keyword table is replaced with a pointer to the Model reference; added a description. All explanatory content and the per-type default tables stay on the page.
- **Vector & geography fields**: the two twin pages are merged into one, preserving both intros, all field types, and the warnings. The old pages are deleted and inbound links updated (the type anchors are unchanged).
- **Custom serialization**: added frontmatter; fixed a stray import (`relation_test_shared` to `my_project_shared`), an incomplete code sample, and a fluff word.
- **Inheritance & polymorphism** and **Backward compatibility**: cross-linked to each other; conformance.
- **Shared packages**: refreshed the version pins to the current 4.0-beta line (Dart `^3.10.3`, Serverpod `4.0.0-beta.0`).

## Notes

- Version values track the current dev line (4.0-beta on `main`); the Dart SDK `^3.10.3` matches the framework template.
- Structural changes (grouping pages into a sidebar category, the section rename, redirects) are deferred to the final PR of this series, so this PR keeps every page at its current URL.
- The `serverpod generate` references on these pages are accurate (the command still exists); their framing is modernized separately in #538, a dedicated cross-cutting pass so every page stays consistent.

Issue #671, opened 2026-07-06
Phase 2: restructure the Concepts (Reference) docs


developerjamiu:
Restructures the Concepts section (public label "Reference") into a predictable, findable shape: eight journey-ordered groups, plus a root-level Reference section for true lookups. Legacy content moves to Upgrading/Archive, and duplicated facts get a single home. It ships as seven reader-coherent PRs, and the section rename to "Concepts" happens last.

## Acceptance bar (every PR)

Each PR is done only when every page it touches, including pages that mostly stay put, passes:

- **Template conformance:** capability pages have a substantive Overview plus action-shaped Operations; reference pages are terse and complete with stable anchors; procedural pages have action-shaped steps, prerequisites, and a verifiable end state.
- **Flow:** reads top-to-bottom as a first-timer, with no forward references.
- **Self-contained:** every term is explained on first mention; no reliance on an adjacent page having been read.
- **Gap coverage:** folds in the relevant open issues and correctness fixes for the area it touches; anything not folded is filed and named in the PR.
- **No information loss:** every unit is preserved or explicitly archived; the build and links pass.

## PRs

- [x] PR 1: Server fundamentals, the reference spine, and the archive scaffold (#664, #538, #565)
- [x] PR 2: Endpoints and APIs (#648, #367, #654)
- [x] PR 3: Data and the database (#652, #657, #659, #504)
- [x] PR 4: Authentication, plus a migrate-from-legacy-auth guide (#503, #507, #508)
- [x] PR 5: The web server and scheduling (#365, #366, #655)
- [x] PR 6: Operations and testing (#629)
- [x] PR 7: Rename to "Concepts", promote the Reference section, redirects and sidebar (#587)

## Notes

- The CLI reference is now auto-generated (#5390), so these PRs link to it rather than rebuild it.
- Items that need framework-team confirmation are deferred to a later pass so they do not block this work: database seeding (#651), server role behavior, and the passkey UI. Structural work around them still ships.

Issue #651, opened 2026-06-24
Document database seeding (insert initial data at startup)


developerjamiu:
## Demand

Kapa "Secrets & Seeding" cluster — 15 conversations / 15 users (highest user count in the audit). Seeding is essentially undocumented.

## Problem

Users repeatedly ask how to insert initial or bulk data at server startup ("feed table data when the server starts", "seed films into the DB", "seed without raw SQL", "seed dummy data"). The only existing mention is an aside about a SQL `INSERT` inside a migration (`06-database/11-migrations.md`), which is not the programmatic pattern users are asking for.

## Scope

- New page under `docs/06-concepts/06-database/` (e.g. `seeding.md`).
- Startup hook / initialization pattern, including "insert only if the table is empty on first run".
- Bulk insert via generated models (not raw SQL).
- Guidance on migrations vs startup seeding, and when to use each.

## Verify before writing

- The recommended startup entry point for running seed code (confirm against the framework source / team).

PR #635, opened 2026-06-17
docs: Make authentication overview headings descriptive

Zfinix:
Makes the first heading on each top-level authentication page more descriptive so each page reads clearly out of context (for example, "Setup" becomes "Set up the authentication module"), and adds a `description` to the frontmatter of every page. Where a heading was lengthened, a `sidebar_label` keeps the sidebar name short.

Pages updated: Get started, Setup, The basics, Working with users, UI components, and Custom overrides.

First of four PRs splitting the same heading and description change across the authentication section. The providers, token managers, and legacy sections follow in separate PRs once this one is reviewed.

Part of #610

PR #676, opened 2026-07-06
docs: Reshape Database concepts, fold ID strategy, many-to-many, and migration gaps

### LINKED ISSUE #652: Expand many-to-many docs: examples, why/when, and troubleshooting
## Demand

Kapa "Many-to-Many Relations" cluster — 15 conversations / 13 users.

## Problem

`06-database/03-relations/03-many-to-many.md` shows how to define the junction model, but:

- No end-to-end "add an entry" example (the most repeated question) — creating and inserting a junction row.
- No troubleshooting for the errors users actually hit ("no named relation 'x'", "no getter personId", `!persist` scope errors), or an explanation of the auto-generated foreign-key field names.
- It shows **how** to do a many-to-many relation, but not **why** or **when** you need one.

Review feedback: the current examples are the bare minimum. We should do an oversight pass on whether to add more examples covering common many-to-many use cases.

## Scope

- "Managing junction records" section: create + `insertRow` a junction row, and the `attach`/`attachRow` path (cross-link `07-relation-queries.md`).
- "Common errors" subsection: auto-generated FK field names, matching `name=` on both sides, field scopes.
- Add **why / when** to reach for a many-to-many relation.
- Oversight pass: add examples for common use cases.

## Note

The why/when explanation may be better owned by a guide once the Phase 2 guides/reference split lands — keep that split in mind so this does not get duplicated.

### LINKED ISSUE #657: Add ID-strategy guidance: int vs UUID, autoincrement, client-set IDs
## Demand

Kapa "User ID & UUID Management" cluster — 9 conversations / 6 users.

## Problem

`06-database/02-models.md` covers id type customization, but does not:

- Give int-vs-UUID decision guidance (performance / storage / use case).
- State plainly that `serial` means auto-increment, or explain `serverOnly` + autoincrement.
- Cover setting a UUID from the client (and the security implications).
- Show having both a UUID id and a secondary serial int on one model.

## Scope

- A "Choosing an ID strategy" section in `06-database/02-models.md` covering the above, with an example model carrying both a UUID id and a serial int.

### LINKED ISSUE #659: Improve migrations docs: custom SQL, Serverpod Cloud apply guidance, troubleshooting
## Demand

Kapa "Applying/Running Migrations" + "Database Migrations" clusters — ~12 conversations combined.

## Problem

`06-database/11-migrations.md` documents `--apply-migrations`, `--role maintenance`, and repair migrations, but:

- Custom SQL migrations are mentioned without structure, example, or when-to-use guidance.
- How migrations are applied on Serverpod Cloud (automatically or manually in the managed environment) is unclear.
- Common mistakes are not addressed: the `--apply-migration` (singular) error; "removing migrations" actually means rolling back via a repair migration, not deletion.

## Scope

- A "Writing custom migrations" section (file structure, when/why, data migrations).
- Serverpod Cloud apply guidance.
- A "Troubleshooting" subsection.

developerjamiu:
Part of the Concepts Phase 2 rollout (#671). Reshapes the Database concept group and folds in three open gaps.

## Folds
- **#657** — adds a "Choosing an ID strategy" section to Database > Models: `int`-serial vs `UuidValue` trade-offs, `serial` as auto-increment, and the security implications of client-set UUIDs.
- **#652** — expands Many-to-many: when to reach for a junction table, "Managing junction records" (`insertRow` plus the `attach`/`detach` path), and a "Common errors" section.
- **#659** — improves Migrations: a "Writing custom SQL" section, Serverpod Cloud apply guidance, and a "Troubleshooting" section.

Advances #504 (database docs epic).

## Conformance
- Noun-first, unquoted `description` frontmatter across all 21 pages in the group (#635).
- House-style sweep on touched pages (banned vocab, backtick-start openings).

## Not included
- **#651 (seeding)** — deferred; needs framework confirmation on the recommended startup entry point.
- One #657 sub-example (a model carrying both a `UuidValue` id and a secondary serial `int`) is omitted: `serial` appears to be the primary-key default only, with no evidence it applies to a non-id field. Worth a framework-team check before documenting.
- Remaining `serverpod generate` references in the group are left for the dedicated modernization pass (#538).

## Notes for review
- Stacked on the Model reference PR (#674) — the base is `concepts-phase2-pr3a-models`, so the diff shows only this group's changes. Retarget to `main` once #674 merges.
- All technical claims verified against the framework `main` checkout; these are stable APIs, not 4.0-only behavior.
- Build passes locally.

Closes #652
Closes #657
Closes #659


docs/06-concepts/06-database/02-models.md
```diff
@@ -167,6 +179,8 @@ fields:
 
 When using `defaultModel=random`, the UUID will be generated when the object is created. Since an id is always assigned the `id` field can be non-nullable.
 
+To have the server assign the id instead of the client, use `defaultPersist=random` without `defaultModel`, so the value is set on insert rather than on the client.
```
Zfinix: It's the database that assigns the id here, not the server, and the server/client contrast doesn't quite hold since `defaultModel` also applies when the object is created on the server. Suggestion:
```suggestion
To have the database assign the id on insert instead of generating it when the object is created, use `defaultPersist=random` without `defaultModel`.
```

Zfinix:
Approving pending resolutions.

Zfinix (approved):

PR #681, opened 2026-07-09
docs: Require nullable for scoped model fields and fix example

developerjamiu:
The `SelectivelyHiddenClass` example on the models page declares a `scope=serverOnly` field with a non-nullable type, which fails code generation. On a non-`serverOnly` class, any field with a scope other than `all` (`serverOnly` or `none`) must be nullable — enforced by the model analyzer since Serverpod 1.2.3.

- Fix the example: `hiddenSecretKey: String` → `String?`.
- Document the rule: a field with a scope other than `all` must be nullable.

Reported in the docs channel (the published example does not generate). Scoped to the current dev version; the same example exists in the versioned docs and can be backported separately if wanted.

### DIFF
--- a/docs/06-concepts/02-models/01-models.md
+++ b/docs/06-concepts/02-models/01-models.md
@@ -53,7 +53,7 @@ fields:
   hiddenSecretKey: String
 ```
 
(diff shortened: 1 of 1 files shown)


Zfinix (approved):
So this has been in the docs since 2.0 right

developerjamiu:
> So this has been in the docs since 2.0 right

Yes

PR #687, opened 2026-07-15
docs: Add project anatomy page and deepen the Server fundamentals section

developerjamiu:
Deep review and rework of the Server fundamentals section, with every technical claim verified against the Serverpod 4.0.0-beta.0 source and a project created with the 4.0.0-beta.0 CLI.

## New page

- **Your Serverpod project**: the section opener explaining what `serverpod create` generates. Starts with the three-package mental model and a short tree, with the full annotated workspace tree in a collapsible. Covers package relations (pub workspace, dependency direction), where user code lives vs generated code, the server entry point (`bin/main.dart`, `run()`, `pod.start()`), the three servers and their ports, the config and migrations directories, the `.serverpod` embedded-database directory, and the editor/agent files.

## Reworked pages

- **Running your server**: the dev loop now includes the database leg (embedded PostgreSQL). Adds degraded-start recovery, the app launch panel (Ctrl+R), `--docker`, single-instance behavior, run-mode gating of app launches, and new "Run the server directly" and "Choose a server role" sections (monolith / serverless / maintenance semantics). Documents the key notation precisely: shortcuts are unshifted presses, Shift variants are written out, and Shift+M / Shift+P force a migration even past data-loss warnings. Notes that P is not shown in the bottom bar.
- **Configuration**: fixes the source-order contradiction in the overview and the `websocketPingInterval` example nesting (it is a top-level key; the nested form was silently ignored). Adds a Run modes section, embedded PostgreSQL under Database backends, the SQLite session-log limitation, dedicated secret environment variables with their precedence, and aligns the example with the scaffolded `config/development.yaml`. Secrets are restructured example-first.
- **Modules**: corrects the module reference example (`AuthUser` lives in `serverpod_auth_core`; references resolve by effective nickname only, and an override replaces the default), pins beta versions exactly, and rewrites the overview reader-first.

## Reference and consistency fixes

- Configuration reference: corrected `consoleEnabled`/`consoleLogFormat` defaults, new Dart-only options (`healthCheckInterval`, `experimentalDiagnosticHandlerTimeout`), and password environment variables.
- Working with endpoints: generation guidance now leads with `serverpod start`'s automatic generation; `serverpod generate` remains as the no-session path.
- Logging: corrected the console-logging default and retargeted a stale pointer; same pointer fix in Streaming and Scheduling configuration.
- How it works: links to the new anatomy page.


docs/06-concepts/lookups/configuration-reference.md
```diff
@@ -59,15 +59,35 @@ Ports, hosts, and connection settings for the API, Insights, and web servers, th
...
+| healthCheckInterval                  | 1m      | How often the server collects health metrics. Set to zero to disable health checks.                    |
+| experimentalDiagnosticHandlerTimeout | 30s     | The timeout for [diagnostic event handlers](../operations/experimental-features#exception-monitoring). |
+
+### Password environment variables
+
+Secrets are read from `config/passwords.yaml` and can be overridden per secret through environment variables; see [Manage secrets](../server-fundamentals/configuration#manage-secrets). Two forms exist, and when both are set for the same secret, the `SERVERPOD_PASSWORD_*` form wins:
+
+| Environment variable            | Overrides passwords-file key                                                     |
+| ------------------------------- | -------------------------------------------------------------------------------- |
+| SERVERPOD*PASSWORD*&lt;name&gt; | Any secret; the prefix is stripped (`SERVERPOD_PASSWORD_database` → `database`). |
```
Zfinix: The asterisks here render as emphasis: `*PASSWORD*` italicizes and the `*` delimiters get dropped, so the cell displays as `SERVERPODPASSWORD<name>` with the underscores gone (confirmed via micromark). The sibling rows avoid this by using plain underscores. Backticking renders the literal env var and matches `SERVERPOD_PASSWORD_database` in the same cell:

```suggestion
| `SERVERPOD_PASSWORD_<name>` | Any secret; the prefix is stripped (`SERVERPOD_PASSWORD_database` → `database`). |
```

Zfinix (approved):
LGTM 🚀 approving with the suggested change applied.

PR #688, opened 2026-07-15
docs: Add a Serverpod Cloud database backups page

developerjamiu:
## Summary

Adds a Serverpod Cloud **Database backups** concept page documenting the new `scloud db backup` and `scloud db schedule` commands: manual snapshots, scheduled backups, restore, plan gating, and pricing.

- New page: `cloud_docs/concepts/database-backups.md` (Overview, when-to-use, the five operations, Pricing, Limits, Related).
- Reconciles the outdated 2-line `## Backups` stub on the Database page (it claimed backups were automatic and needed no configuration) into a short pointer to the new page.
- Renumbers two sidebar positions so backups sits directly after Database (`personal-access-tokens` 6→7, `deployment-hooks` 7→8).

## Do not merge until the feature is released

The backup CLI is on `main` in the cloud repo but not yet in a released `scloud`, and the web console is still in progress (CLD-505). Keep this open for review and merge once the feature ships (targeting ~week of Jul 20). Console sections will be added in a follow-up once the console lands.

## Verification

Every claim was checked against the cloud repo at `origin/main` (backup CLI shipped in #2090):

- CLI surface, flags, defaults, and output/confirmation wording read from the command source, not written from memory.
- Growth-plan gating and the split of gated vs. any-plan commands confirmed in the databases domain/endpoint code.
- Pricing ($6 / 10 GB, separate metered charge) and the 100 manual-snapshot cap confirmed in the billing and policy config.
- `npm run build` passes (no broken links or anchors).

## Notes for reviewers

- **Downgrade with backups** is documented as if enforced (a Growth→smaller-plan change is blocked while backups exist). This behavior is intended but not yet implemented; documenting it as in effect was agreed with the platform team. Worth a final confirm before merge.
- **Scheduled-snapshot retention default** is described as "the platform's default retention" rather than a specific number, matching the CLI's own wording. If we want to state the exact value, that's the one detail still to pin.
- **Pricing rate** is stated on the page and linked to the plans page. Once a dedicated plans/pricing page exists, the exact rate can move there and this page can link out.
- The provider is deliberately never named; backups are described as a feature of the managed database.

### DIFF
--- a/cloud_docs/concepts/database-backups.md
+++ b/cloud_docs/concepts/database-backups.md
@@ -6,10 +6,10 @@ description: Database backups on Serverpod Cloud are point-in-time snapshots of 
 
 # Database backups
 
-A bad migration or an accidental bulk delete can lose data you needed. Serverpod Cloud can take point-in-time snapshots of your managed database, on demand or on a recurring schedule, and restore the whole database to any snapshot in one step. Backups run on the platform, so there is nothing for you to host or maintain. Today you manage them with the `scloud` CLI, and support in the web console is coming.
+A bad migration or an accidental bulk delete can lose data you needed. Serverpod Cloud can take point-in-time snapshots of your managed database, on demand or on a recurring schedule, and restore the whole database to any snapshot in one step. Backups run on the platform, so there is nothing for you to host or maintain. You manage them with the `scloud` CLI.
 
 :::info
-Backups are available on the **Growth** plan. On other plans, creating a snapshot, setting a schedule, and restoring return an error that backups aren't available for the project's plan. Listing snapshots, deleting snapshots, and viewing or disabling the schedule work on any plan, so a project on a smaller plan can still see and clean up backups it already has.
+Backups are a **Growth** plan feature. On Growth you can create snapshots, schedule backups, and restore. Listing snapshots, deleting snapshots, and viewing or disabling the schedule are available on every plan, so a project on a smaller plan can still see and clean up backups it already has.
 :::
 
 ## When to use each kind of backup
@@ -87,7 +87,7 @@ The current state is retained by the provider as a separate backup.
 Do you want to proceed?
 ```
 
-The connection string and database credentials do not change, so your deployed server keeps working against the same database with no config change. The state from just before the restore is kept as a separate backup, so an accidental restore can itself be recovered from.
+The connection string and database credentials do not change, so your deployed server keeps working against the same database with no config change.
 
 ## Delete a snapshot
 
@@ -101,7 +101,7 @@ Cloud asks for confirmation, then deletes the snapshot permanently. Scheduled sn
 
 ## Pricing
 
-Backup storage is billed separately from your regular database storage, as its own charge rather than counting against your database usage. It uses the same package pricing as database storage, $6 per 10 GB per month, prorated for partial months, and applies only on the Growth plan. A schedule usually costs less than keeping the same number of manual snapshots. Each manual snapshot is a full copy, while a scheduled series stores one full copy plus the changes between snapshots. For current plan pricing, see [Serverpod Cloud plans](https://serverpod.dev/cloud).
+Backup storage is billed separately from your regular database storage, as its own metered charge rather than counting against your database usage. It is priced the same as database storage, billed by usage and prorated for partial months, and applies only on the Growth plan. A schedule usually costs less than keeping the same number of manual snapshots. Each manual snapshot is a full copy, while a scheduled series stores one full copy plus the changes between snapshots. For current plan pricing, see [Serverpod Cloud plans](https://serverpod.dev/cloud).
 
 ## Limits
 
--- a/cloud_docs/concepts/database-backups.md
+++ b/cloud_docs/concepts/database-backups.md
@@ -27,7 +27,7 @@ A schedule takes snapshots for you at a fixed frequency. Each database has one s
 scloud db schedule set --frequency weekly --day 1 --hour 3 --retention 30d
 ```
 
-Use `--day` to pick the day of the week (1-7) for a weekly schedule or the day of the month (1-31) for a monthly one; it defaults to 1 and does not apply to a daily schedule. Use `--hour` for the hour of the day (0-23, in UTC); it defaults to 0. Use `--retention` (for example `30d`) to set how long each scheduled snapshot is kept before it is deleted automatically. Without `--retention`, scheduled snapshots are kept for the platform's default retention.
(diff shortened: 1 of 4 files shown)


Zfinix (approved):
LGTM 🚀 great work on this.

nit (non-blocking): the `$6 / 10 GB` rate will drift as plans change — consider dropping the exact number here and letting the plans page own it, keeping the "separate metered charge, same rate as database storage" framing.

cloud_docs/concepts/database-backups.md
```diff
@@ -0,0 +1,116 @@
...
+
+```bash
+scloud db backup delete snap-3f2a9c
+```
+
+Cloud asks for confirmation, then deletes the snapshot permanently. Scheduled snapshots are removed automatically once they pass the schedule's retention, so you mostly delete manual snapshots this way.
+
+## Pricing
+
+Backup storage is billed separately from your regular database storage, as its own charge rather than counting against your database usage. It uses the same package pricing as database storage, $6 per 10 GB per month, prorated for partial months, and applies only on the Growth plan. A schedule usually costs less than keeping the same number of manual snapshots. Each manual snapshot is a full copy, while a scheduled series stores one full copy plus the changes between snapshots. For current plan pricing, see [Serverpod Cloud plans](https://serverpod.dev/cloud).
```
Zfinix: nit (non-blocking): the `$6 / 10 GB` rate will drift as plans change. Consider dropping the exact number here and letting the plans page own it, keeping the "separate metered charge, same rate as database storage" framing.

cloud_docs/concepts/database-backups.md
```diff
@@ -0,0 +1,116 @@
+---
+title: Database backups
+sidebar_position: 6
+description: Database backups on Serverpod Cloud are point-in-time snapshots of the managed database, taken on demand or on a schedule and restorable in one step.
+---
+
+# Database backups
+
+A bad migration or an accidental bulk delete can lose data you needed. Serverpod Cloud can take point-in-time snapshots of your managed database, on demand or on a recurring schedule, and restore the whole database to any snapshot in one step. Backups run on the platform, so there is nothing for you to host or maintain. Today you manage them with the `scloud` CLI, and support in the web console is coming.
```
Isakdl: Suggestion: Don't mention unsupported platforms, it just adds unnecessary information. Instead always guide the user to what works today! 

With that said this should be updated with the console information before merge as that is the last part to be completed before we consider this feature ready.

cloud_docs/concepts/database-backups.md
```diff
@@ -0,0 +1,116 @@
...
+
+:::info
+Backups are available on the **Growth** plan. On other plans, creating a snapshot, setting a schedule, and restoring return an error that backups aren't available for the project's plan. Listing snapshots, deleting snapshots, and viewing or disabling the schedule work on any plan, so a project on a smaller plan can still see and clean up backups it already has.
```
Isakdl: Suggestion: Don't mention "error" etc. Use positive language about where a feature is available.

cloud_docs/concepts/database-backups.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Set up automatic backups
+
+A schedule takes snapshots for you at a fixed frequency. Each database has one schedule. Set it with `scloud db schedule set` and a frequency of `daily`, `weekly`, or `monthly`:
+
+```bash
+scloud db schedule set --frequency weekly --day 1 --hour 3 --retention 30d
+```
+
+Use `--day` to pick the day of the week (1-7) for a weekly schedule or the day of the month (1-31) for a monthly one; it defaults to 1 and does not apply to a daily schedule. Use `--hour` for the hour of the day (0-23, in UTC); it defaults to 0. Use `--retention` (for example `30d`) to set how long each scheduled snapshot is kept before it is deleted automatically. Without `--retention`, scheduled snapshots are kept for the platform's default retention.
```
Isakdl: > snapshots are kept for the platform's default retention.

Suggestion: This should specify what the default is not "platform's default" that will not help anyone reading this.

cloud_docs/concepts/database-backups.md
```diff
@@ -0,0 +1,116 @@
...
+
+```text
+WARNING: Restores the database for project "my-app" to snapshot "snap-3f2a9c".
+The live database is replaced with the data from the snapshot.
+The current state is retained by the provider as a separate backup.
+
+Do you want to proceed?
+```
+
+The connection string and database credentials do not change, so your deployed server keeps working against the same database with no config change. The state from just before the restore is kept as a separate backup, so an accidental restore can itself be recovered from.
```
Isakdl: Comment: This is true, but we have no mechanism exposed _right now_ to do the recover. Sync with @christerswahn what will be available, but my hunch is that we should remove this from the docs until we have added this support.

cloud_docs/concepts/database-backups.md
Isakdl: Agreed!

Isakdl (dismissed):
Good work! I like it,

I do have some thoughts and suggestions. I will probably not be available when you will merge this, so you can dismiss my needed rereview.

cloud_docs/concepts/database-backups.md
developerjamiu: What is the default value here @Isakdl @christerswahn?

cloud_docs/concepts/database-backups.md
christerswahn: I agree, remove the second sentence until we have exposed support for it.

cloud_docs/concepts/database-backups.md
christerswahn: According to the current code in Ground Control, default snapshot retention is 24h.

christerswahn (commented):
Added responses to a couple of comments.

cloud_docs/concepts/database-backups.md
developerjamiu: Applied. The paragraph keeps the same-rate framing and the plans page owns the number.

cloud_docs/concepts/database-backups.md
developerjamiu: Set both mentions to 24 hours.

cloud_docs/concepts/database-backups.md
developerjamiu: Removed in the last push. Note the `scloud` restore confirmation still prints "the current state is retained...", so fully removing the promise needs that CLI message updated too.

christerswahn (approved):
LGTM

PR #693, opened 2026-07-17
docs: Deepen the Endpoints and APIs section with verified 4.0 behavior

developerjamiu:
Deep review and rework of the Endpoints & APIs section. Every technical claim was verified against the Serverpod 4.0.0-beta.0 source, a project created with the 4.0.0-beta.0 CLI, and, for the new examples, integration tests against a running server.

## Structure

- Pages reordered essentials-first: endpoints, sessions, error handling, streaming, server events, file uploads, then inheritance, middleware, and HTTP configuration. URLs are unchanged (numeric prefixes are stripped).
- The section index now opens with a short map and closes with a Related block covering every sibling page, and gains a "Restrict access to your endpoints" summary pointing to Authentication.

## Corrections

- **Client setup now matches the 4.0 scaffold**: `getServerUrl()` with the `SERVER_URL` define, `assets/config.json`, and platform-aware localhost. The previous guidance taught a hand-rolled pattern with a different variable name than the framework reads.
- **Sessions**: two claims removed as incorrect in 4.0 (use-after-close throwing a `StateError`; logs accumulating in memory until close), the deprecated `futureCallWithDelay` replaced with the typed `futureCalls` builder, and the non-compiling testing example replaced with a pointer to the testing docs. `withSession` is now the primary manual-session pattern.
- **Streaming**: the error-handling examples could not compile (`SerializableException` is abstract); replaced with a generated-exception example validated by integration tests in both directions. An inverted inStream/outStream sentence fixed.
- **Server events**: the quick-reference named a method that does not exist; the real `authenticationRevoked` is now documented, and `postMessage` is awaited in examples.
- **File uploads**: corrected the S3 secret key names in prose, replaced root-user credential advice with an IAM user, removed a stale Terraform claim (no such scaffolding ships in 4.0), fixed a server/client snippet mismatch, and corrected the stream-memory claim (a multipart upload without a length buffers the file).
- Terminology: "streaming methods" consistently names the current API (two pages used the deprecated API's name for it).

## Additions (all verified in source)

- The eight client-side `MethodStreamException` types with when each is thrown.
- `sendAsRaw` and `logSessions` endpoint flags, `session.serverpod`, storage `deleteFile`/`getPublicUrls`/`storeFile(expiration:)`, the upload-progress limitation, the 413 request-size failure mode, and `dynamic` plus vector and geography types in the parameter list.
- An empirically tested rate-limit middleware example, with its production caveats.

## Also touched

- how-it-works and the JWT token manager page: terminology and Future-or-Stream fixes.
- The CLI generate reference partial: two links that silently resolved to the 3.4 versioned pages now stay in the current version.


Zfinix (approved):
LGTM 🚀

Issue #504, opened 2026-04-29
epic: Improve the database docs


Swiftaxe:
### Summary of the issue

Investigate and improve the database docs to:

- Clarify
- Simplify
- Add missing steps and code examples
- Address coverage gaps and common questions

developerjamiu:
The database docs rework shipped in #674 (model reference), #676 (database concepts), and #696 (the full section pass). Seeding stays tracked in #651.

PR #696, opened 2026-07-21
docs: Rework the Data & the database section for 4.0

developerjamiu:
Section-by-section deep rework of the Concepts area, part 3 of the series (after Server fundamentals in #687 and Endpoints & APIs in #693). This pass covers the Data & the database section: 22 pages across the models and database subgroups, plus the pages they touch.

## Structure

- **Vector & geography fields** moved from the models subgroup into the database subgroup, next to the query pages that use it.
- **Backward compatibility** moved to Endpoints & APIs, where its rules (endpoint signatures, future calls) belong.
- Database pages renumbered essentials-first: tables → CRUD → filtering → sorting → pagination → relations → relation queries → vector & geography → indexing → transactions → row locking → raw access → migrations → connection → runtime parameters → client-side database → exceptions.
- New **Relations landing page**: the category label now opens a short orientation page with a "which shape when" table.
- All inbound links updated; URL redirects for the two moved pages are queued for the 4.0 GA cutover (the old URLs only ever existed under /next).

## Content highlights

- **New coverage for 4.0 surface**: upsert (`upsertRow`/`upsert` with `conflictColumns`, `updateColumns`, `updateWhere`, `noReturn`), `dynamic` model fields, `DatabaseResult`, `databaseInterceptor`, `Transaction.cancel()`, the `tail` keyword, `Uri`/`BigInt` default values, the client-migrations pipeline, and a wire-format table for the built-in types.
- **Corrections**: the `--force`/`--empty` conflation on migrations, the `repair-migration` directory name, a non-compiling filtering example (Dart operator precedence), the comparison-operator type list (String and UuidValue were missing), the inheritance table rule (only an ancestor with a table blocks a child; sibling branches may each define one), future-call compatibility rules rewritten around the real API, and the modules relation page rebuilt on `serverpod_auth_core`.
- **SQLite behavior** documented where it diverges: `like`/`ilike` case rules, ignored isolation levels, row locking, vector/geography limits, and the simple-query methods.
- **Client-side database** page rebuilt: enabling the `database` keyword, session lifecycle with `close()`, on-device migrations with `runMigrations`/`isDebugMode`, and a gathered list of SQLite differences.
- **Model reference** gained rows for `database`, `serializationDataType`, `column`, `operatorClass`, and `tail`, and corrected the `immutable`, `sealed`, `default`, `defaultModel`, and `unique` rows against the CLI validators.

Every new technical claim was verified against the framework at 4.0.0-beta.0, and the upsert, cursor-pagination, and raw-access examples were run against a real project (12/12 tests green). Build passes with no broken links or anchors.

Advances #504.


docs/06-concepts/03-data-and-the-database/02-database/03-filtering.md
```diff
@@ -49,38 +47,15 @@ await User.db.find(
...
-
-In the example we fetch all users that are 25 years old or younger.
+:::info
+Comparison operators check value types at runtime. Comparing a `double` column against an `int` literal, such as `t.price > 500`, compiles but throws. Write the literal to match the column type, `t.price > 500.0`.
+:::
 
 ### Between
 
-The between method takes two values and checks if the columns value is between the two input variables _inclusively_.
+The between method takes two values and checks if the column's value is between them, inclusive of the boundaries. It is available on `int`, `double`, `Duration`, and `DateTime` columns.
```
Zfinix, line 58: `between`/`notBetween` come from the same `ColumnComparable` mixin as the comparison operators, so they are also available on `String` and `UuidValue` columns. This four-type list contradicts the six-type `>`/`>=`/`<`/`<=` list a few lines above. Add `String` and `UuidValue` here so the two match.

Zfinix (approved):
LGTM 🚀 Approved once the inline `between` fix is applied and merge conflicts are fixed

docs/06-concepts/03-data-and-the-database/02-database/03-filtering.md
developerjamiu: The four-type list is deliberate. `between`/`notBetween` don't come from `ColumnComparable`: they live in the separate `_ColumnComparisonBetweenOperations` mixin, which only `ColumnInt`, `ColumnDouble`, `ColumnDuration`, and `ColumnDateTime` mix in (columns.dart:238, 253, 312, 327). `ColumnString` and `ColumnUuid` get the comparison operators from `ColumnComparable` but not `between`, so `t.name.between('a', 'z')` doesn't compile. The two lists differ because the API does.

PR #700, opened 2026-07-21
docs: Document the `noReturn` option on write operations

marcelomendoncasoares:
## Summary

The batch and filtered write methods accept `noReturn: true`, which runs the write without reading the affected rows back. The CRUD page had no mention of it.

## Changes

- Add a "Skipping returned rows" section to the CRUD page.
- Show `noReturn: true` on `updateWhere` and explain that the write still runs with the same filters, conflict handling, transaction, and atomicity guarantees.
- Describe the performance tradeoff and when it is worth taking.
- List the methods that accept it: `insert`, `update`, `updateWhere`, `upsert`, `delete`, and `deleteWhere`. The single-row variants always return their row.
- Note that `orderBy` and `orderByList` no longer affect the (empty) result, but still decide which rows `limit` and `offset` select.

## Note on merge order

The method list includes `upsert`, which is documented in the companion PR for the upsert methods. If this PR merges first, `upsert` is named here before it has a section of its own on the page.

## Reference

- serverpod/serverpod#5371
- serverpod/serverpod#5118


developerjamiu (approved):
LGTM

PR #701, opened 2026-07-21
docs: Document the ORM upsert methods

marcelomendoncasoares:
## Summary

`upsertRow` and the batch `upsert` are generated on every table model but were missing from the CRUD page, so there was no documented way to insert-or-update a row.

## Changes

- Add an "Upsert" section to the CRUD page, right after Create and its conflict handling.
- Extend the shared `Company` example model with a `unique` name, `address`, `employeeCount`, and `isActive`, so the conflict examples have something to conflict on. The `address` field was already used by an existing example on the page but missing from the model.
- Show `upsertRow` with `conflictColumns`, and explain the default update set.
- State that the primary `id` column can be a conflict target.
- Document `updateColumns` and `updateWhere`, and explain why `upsertRow` returns a nullable value.
- Document the batch `upsert`: atomicity, transaction support, a result that can be shorter than the input, and `noReturn`.
- Explain how non-persistent fields behave.

## Reference

- serverpod/serverpod#4832
- serverpod/serverpod#232
- serverpod/serverpod#5342, serverpod/serverpod#5353, serverpod/serverpod#5428


developerjamiu (approved):
LGTM

PR #702, opened 2026-07-21
docs: Document database interceptors

marcelomendoncasoares:
## Summary

The `Serverpod` constructor and the `withServerpod` test helper accept a `databaseInterceptor` callback that decides what `Session.db` returns. Nothing in the docs covered it.

## Changes

- Add a "Database interceptors" page at the end of the database section, so it does not interrupt the normal CRUD path.
- Show the callback signature, that it receives the session and the framework database, and that it is called once per session.
- Show returning a custom `Database` implementation and list what it is useful for: tracing, metrics, tenant scoping, policy checks, and safety guards.
- Note that a policy layer has to cover raw access and transactions to be effective.
- Carry over the API stability warning from the implementation: custom layers depend on `serverpod_database`, which can change in a breaking way in a minor release.
- Show the same callback in a `withServerpod` test and add `databaseInterceptor` to the test helper configuration table.
- Cross-link from the raw access page.

## Reference

- serverpod/serverpod#5344
- serverpod/serverpod#5337


developerjamiu (approved):
LGTM

PR #703, opened 2026-07-21
docs: Document sign-in widget localization

### LINKED ISSUE #5125: docs: Missing documentation for localization support in sign-in.
### Summary of the issue

Sign in has support for localization through `SignInLocalizationProvider`. This should be documented, and we should (probably) also mention it in our `serverpod-auth` skill.

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [ ] I want to work on this

marcelomendoncasoares:
Closes: https://github.com/serverpod/serverpod/issues/5125

## Summary

`SignInLocalizationProvider` lets an app replace the English text of the authentication widgets, but nothing in the docs mentioned it.

## Changes

- Add a "Localization" section to the authentication UI components page. It is a top-level section because the provider applies to every authentication widget, not only `SignInWidget`.
- List the text objects and what each one covers before the example.
- Add one complete example translating the shared texts, the email flow, the password requirements, and every provider button with explicit English texts.
- Explain the fallback: anything left out keeps the built-in English text, and `defaults.copyWith` covers partial translations.
- State that the provider does not resolve Flutter's `Locale`, and that the app has to rebuild it when the locale changes.
- Note that the Firebase provider ships a controller rather than a finished UI, so it is not covered by these text objects.
- Fix the sentence case of the existing "Disabling providers" heading. The slug is unchanged.

## Reference

- serverpod/serverpod#4711
- serverpod/serverpod#4496

### DIFF
--- a/docs/06-concepts/04-authentication/06-ui-components.md
+++ b/docs/06-concepts/04-authentication/06-ui-components.md
@@ -52,7 +52,93 @@ class SignInPage extends StatelessWidget {
 }
 ```
 
-### Disabling Providers
+### Localize sign-in text
+
+Wrap `SignInWidget`, or individual authentication widgets, in a `SignInLocalizationProvider` to replace their built-in English text. The following example supplies Brazilian Portuguese text for the common UI, the complete email flow, password requirements, and each supported provider button:
+
+```dart
+const basicTexts = BasicSignInTexts(
+  noAuthenticationProvidersConfigured:
+      'Nenhum provedor de autenticação configurado',
+  orContinueWith: 'ou continue com',
+);
+
+const emailTexts = EmailSignInTexts(
+  title: 'Entrar com e-mail',
+  forgotPassword: 'Esqueceu sua senha?',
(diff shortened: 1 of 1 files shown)


developerjamiu:
The PR description says the example translates into Brazilian Portuguese, but all the strings in the diff are English. The English version reads well as an API reference, so maybe just update the description, or give the copyWith example one non-English string (e.g. signIn: 'Entrar') so the translation point shows.

developerjamiu (approved):
Changes looks good! Please check the added (minor) comment before merge

marcelomendoncasoares:
> The PR description says the example translates into Brazilian Portuguese, but all the strings in the diff are English. The English version reads well as an API reference, so maybe just update the description, or give the copyWith example one non-English string (e.g. signIn: 'Entrar') so the translation point shows.

Updated the stale description. It wasn't even supposed for the PR description to be that long, but the agent went ahead of itself creating the PR without permission, and I forgot to review the PR description after pushing my code review.

PR #704, opened 2026-07-21
docs: Document embedded PostgreSQL

marcelomendoncasoares:
## Summary

Embedded PostgreSQL is the default for the `development` and `test` run modes in new projects, and the configuration page introduces `dataPath`, but nothing explained what the setting actually does, where the files live, or when to move off it. This fills that in on the database connection page.

## Changes

- Replace "Development database" with an "Embedded PostgreSQL" section and move the Docker instructions into their own section.
- Explain what `dataPath` starts, and link to the configuration page rather than repeating it.
- Cover storage: relative path resolution, one directory per run mode, keeping it out of version control, the first-run binary download and its per-user cache location, and the supported platforms.
- Cover lifecycle: startup before the connection pool, the Unix socket transport, attaching to a cluster another process already runs, persistence across restarts, and recovery after an unclean exit.
- Document resetting the database by deleting `dataPath`, with a warning about the data loss.
- State the production boundary and how to move to an externally managed instance.
- Add an advanced section for using `serverpod_embedded_postgres` directly: `start`, transports, connection details, `attach`, `prefetch`, `stop`, `reset`, and the sealed exception type.
- Update the `SERVERPOD_DATABASE_DATA_PATH` row in the configuration reference.

## Reference

- serverpod/serverpod#5090, serverpod/serverpod#5316
- serverpod/serverpod#5087, serverpod/serverpod#5068, serverpod/serverpod#2421


developerjamiu (commented):
The section says each test gets an isolated temporary data directory that is deleted on teardown, but the template pins config/test.yaml to dataPath: .serverpod/test/pgdata, and I can't find a temp-directory substitution in serverpod_test or the config loading. Can you point to the implementation? If tests actually use the configured fixed path, this section and the get-started edit need rewording.

marcelomendoncasoares:
> The section says each test gets an isolated temporary data directory that is deleted on teardown, but the template pins config/test.yaml to dataPath: .serverpod/test/pgdata, and I can't find a temp-directory substitution in serverpod_test or the config loading. Can you point to the implementation? If tests actually use the configured fixed path, this section and the get-started edit need rewording.

Sorry, this was my confusion with one of the approaches that did not end as final. I have updated the docs and removed the changes that would conflict with #721.

docs/06-concepts/03-data-and-the-database/02-database/19-embedded-postgres.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Run integration tests
+
+Integration tests bring their own database, so there is nothing to install or start first:
+
+```bash
+$ dart test
+```
+
+The started database server will be held open until the test process exits. Every suite in a test process shares the same database server, but each `withServerpod` group creates a randomly named database of its own on it that will be dropped on teardown. That is where the isolation comes from: groups run in parallel without seeing each other's data or leaving anything behind between runs.
```
developerjamiu, line 60: **Suggestion:** The isolation attribution isn't right here. `serverpod_test` never creates or drops a database, and the embedded server is created once with `databaseName: config.name` from `config/test.yaml` (`postgres_pool_manager.dart:200`), so there is no per-group randomly named database.

Isolation comes from the rollback transaction each test runs in (`RollbackDatabase.afterEach`), which is what the Testing configuration page documents, so as written the two pages would disagree about the same mechanism.

The rest of the paragraph checks out and is worth keeping: one shared server per process (the pool manager attaches to a running cluster rather than starting a second), a persistent data directory, and data left behind after a hard kill. Suggest dropping just the random-database sentence.

developerjamiu (approved):

docs/06-concepts/03-data-and-the-database/02-database/19-embedded-postgres.md
marcelomendoncasoares: You might have checked against 4.0.0-beta.0, but this changed for 4.0.0-beta.1 (just released yesterday). We now have two isolation layers: the ephemeral per-group database (`sp_test_*`) and the rollback transaction. Both work for the embedded database or for a docker Postgres.

Before the server starts, `TestServerpod` overrides the config to use that ephemeral name (`db.withName(_targetDatabaseName)`), so the pool connects to `sp_test_*`, not the `test.yaml` database.

When the test database is started (`TestServerpod.start()`), it will create an ephemeral database:

https://github.com/serverpod/serverpod/blob/692a23228863f52a2207d3cde195d6cbb31463a5/packages/serverpod_test/lib/src/test_serverpod.dart#L233-L246

The Postgres implementation of the ephemeral database will create an empty database:

https://github.com/serverpod/serverpod/blob/692a23228863f52a2207d3cde195d6cbb31463a5/packages/serverpod_test/lib/src/ephemeral_test_database.dart#L124-L139

https://github.com/serverpod/serverpod/blob/692a23228863f52a2207d3cde195d6cbb31463a5/packages/serverpod_test/lib/src/test_database_manager.dart#L45-L53

When the test database is finished, `tearDownAll` calls `TestServerpod.shutdown()`:

https://github.com/serverpod/serverpod/blob/692a23228863f52a2207d3cde195d6cbb31463a5/packages/serverpod_test/lib/src/test_serverpod.dart#L286-L297

And shutdown will drop the temporary database:

https://github.com/serverpod/serverpod/blob/692a23228863f52a2207d3cde195d6cbb31463a5/packages/serverpod_test/lib/src/ephemeral_test_database.dart#L141-L142

https://github.com/serverpod/serverpod/blob/692a23228863f52a2207d3cde195d6cbb31463a5/packages/serverpod_test/lib/src/test_database_manager.dart#L55-L59

PR #706, opened 2026-07-21
docs: Document exception inheritance and sealed exception hierarchies

marcelomendoncasoares:
## Summary

Exception models support `extends` and `sealed`, which is what makes an exhaustive `switch` over server failures possible in the app. The error handling page only covered flat exceptions, and the model reference listed both keywords as class-only.

## Changes

- Add an "Exception hierarchies" section to the error handling page.
- Show `extends` and `sealed` on exception model definitions.
- Add an endpoint that throws the concrete child exceptions and an app that catches the base type and switches over the subtypes exhaustively.
- Explain what a sealed base buys the app when a new exception is added later.
- Cover hierarchies in shared packages, including the rule that every subtype has to stay in the package declaring the base.
- State the validation rule that exceptions and regular models cannot be mixed in one hierarchy.
- Mark `sealed` as applicable to exceptions in the model reference, adjust the `extends` note, and link the error handling page from Related.

## Reference

- serverpod/serverpod#5307
- serverpod/serverpod#4249


developerjamiu (approved):
LGTM

PR #707, opened 2026-07-21
docs: Document the Serverpod Cloud email identity provider default

marcelomendoncasoares:
## Summary

Newly generated projects register `ServerpodCloudEmailIdpConfig`, not the `EmailIdpConfig` the setup page described. A reader following the page saw a configuration that does not match their `server.dart` and had nothing explaining where the emails actually come from.

## Changes

- Rewrite "Server-side configuration" around the two delivery options, with the generated default first.
- Document `ServerpodCloudEmailIdpConfig`: `appDisplayName`, codes logged in `development` and `test`, and email sent in `staging` and `production`.
- Describe the best-effort delivery behavior, where failures are logged, and why registration and password-reset responses do not change when delivery fails.
- Keep the custom provider path as `EmailIdpConfigFromPasswords` with the `mailer` example, and move the pepper into the parameter list where it now comes from.
- Move the endpoint and migration step into its own subsection. It applies to both configurations, but was previously nested inside the custom provider branch.

## Reference

- serverpod/serverpod#5411
- serverpod/serverpod#5376


docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
```diff
@@ -5,15 +5,52 @@ description: Sign in with Email lets users authenticate with an email and passwo
...
+
+Then extend the abstract endpoint to expose the email authentication routes on the server. Create the file anywhere under your server's `lib/` directory (for example, `<project>_server/lib/src/endpoints/`); the generator picks it up:
+
+```dart
+import 'package:serverpod_auth_idp_server/providers/email.dart';
+
+class EmailIdpEndpoint extends EmailIdpBaseEndpoint {}
+```
+
+Then, start the server with `serverpod start` to generate the client code, then create and apply the migration that initializes the database for the provider (in the `serverpod start` terminal, press **M**, then **A**). More detailed instructions can be found in the general [identity providers setup section](../../setup#identity-providers-configuration).
```
developerjamiu: **Suggestion:** The endpoint and migration steps are already done in newly generated projects: `serverpod create` runs with auth enabled, scaffolds `lib/src/auth/email_idp_endpoint.dart` with the `EmailIdpBaseEndpoint` extension, and creates the initial migration. As written, a new-project reader who was just told "Newly generated projects already configure the email identity provider" is then instructed to create a file they already have. Suggest scoping these steps the same way, e.g.: "Newly generated projects already include this endpoint at `lib/src/auth/email_idp_endpoint.dart`. If you are adding the auth module to an existing project, extend the abstract endpoint and create and apply the migration yourself:".

docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
```diff
@@ -5,15 +5,52 @@ description: Sign in with Email lets users authenticate with an email and passwo
...
+      appDisplayName: 'My App',
+    ),
+  ],
+);
+```
+
+Set `appDisplayName` to the name recipients should see in the verification emails. In the `development` and `test` run modes the codes are written to the server log instead of being sent, so you can complete the flow locally. Once the app is deployed to Serverpod Cloud, they are sent as email.
+
+:::info
+`ServerpodCloudEmailIdpConfig` sends email only for apps running on Serverpod Cloud, which injects the key it authenticates with on deploy. If you host the server yourself, switch to [your own email provider](#use-your-own-email-provider) before you go to staging or production.
```
developerjamiu: **Question:** The PR description mentions adding a paragraph to the Cloud passwords page describing `scloudAuthEmailKey`, linked from here, but the diff only touches the setup page and the key is never named in it. The info box's "the key it authenticates with" is left dangling as a result. Was the passwords-page paragraph dropped deliberately, or did that change not make it into the push?

docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
```diff
@@ -5,15 +5,52 @@ description: Sign in with Email lets users authenticate with an email and passwo
...
+
+If a code cannot be sent, the failure is recorded in the session log and the sign-in flow continues unchanged, so check your [server logs](../../../operations/logging) when a user reports a missing code.
+
+### Use your own email provider
+
+Serverpod Cloud delivery is there to get sign-in working quickly, and it sends a standard message carrying your `appDisplayName`. You might prefer using a custom email provider to have full control over the body, layout, and language of the emails. For servers hosted outside of Serverpod Cloud, it is the only option.
+
+Changing the email provider is done by replacing `ServerpodCloudEmailIdpConfig` with `EmailIdpConfig`, which requires you to pass your own callbacks for the two codes. One convenient option is the [mailer](https://pub.dev/packages/mailer) package, which can send emails through any SMTP service. Most email providers, such as Resend, Sendgrid or Mandrill, support SMTP.
```
developerjamiu: **Question:** The PR description says the custom path stays on `EmailIdpConfigFromPasswords`, but the example uses `EmailIdpConfig` with `secretHashPepper: pod.getPassword('emailSecretHashPepper')!`. The class docs on `ServerpodCloudEmailIdpConfig` point custom-provider users to `EmailIdpConfigFromPasswords`, which reads the same pepper itself and shortens the example to the two callbacks. Which one is intended here? If `EmailIdpConfigFromPasswords`, the Basic configuration options list needs a matching pass, since it documents `secretHashPepper` as required.

docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
```diff
@@ -5,15 +5,52 @@ description: Sign in with Email lets users authenticate with an email and passwo
...
+  ],
+  identityProviderBuilders: [
+    ServerpodCloudEmailIdpConfig(
+      appDisplayName: 'My App',
+    ),
+  ],
+);
+```
+
+Set `appDisplayName` to the name recipients should see in the verification emails. In the `development` and `test` run modes the codes are written to the server log instead of being sent, so you can complete the flow locally. Once the app is deployed to Serverpod Cloud, they are sent as email.
```
developerjamiu: **Suggestion:** Sending is keyed to the run mode, not to where the app runs (the source switches on `staging`/`production`, and the PR description says the same). A self-hosted server in production run mode doesn't log the codes; it attempts to send and the failure is only logged.

```suggestion
Set `appDisplayName` to the name recipients should see in the verification emails. In the `development` and `test` run modes the codes are written to the server log instead of being sent, so you can complete the flow locally. In the `staging` and `production` run modes, the codes are sent as email through the Serverpod Cloud email service.
```

docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
```diff
@@ -5,15 +5,52 @@ description: Sign in with Email lets users authenticate with an email and passwo
 
 # Set up email sign-in
 
-To properly configure Sign in with Email, you must connect your Serverpod to an external service that can send the emails. One convenient option is the [mailer](https://pub.dev/packages/mailer) package, which can send emails through any SMTP service. Most email providers, such as Resend, Sendgrid or Mandrill, support SMTP.
+Sign in with Email verifies the user's address with a code, both when they register and when they reset their password.
```
developerjamiu, line 8: **Suggestion:** The frontmatter description still says "Connect Serverpod to an SMTP service and configure the email identity provider", which now contradicts this intro, since the default needs no SMTP service. Something like: "Sign in with Email lets users authenticate with an email and password. Email delivery works out of the box on Serverpod Cloud, or through your own provider."

docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
marcelomendoncasoares: I dropped it intentionally and forgot to update the PR description. The Cloud docs already mention a generic `Auth` group, so no need to detail this secret specifically. Updated the PR description now.

docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
marcelomendoncasoares: The option to maintain the original instead of the `*FromPasswords` was deliberate to match the pattern in which these docs were written: explicit configuration, with the convenience alternative shown below. But I see you have inverted this pattern on the published guides for Google and Apple, so we can invert it here as well to keep it consistent. I'll favor the `EmailConfigFromPasswords`.

docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
marcelomendoncasoares: Addressed at https://github.com/serverpod/serverpod_docs/pull/707/commits/f17bdee8efd1dad3c480705104cae3cd478a7f98.

docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
marcelomendoncasoares: Addressed at https://github.com/serverpod/serverpod_docs/pull/707/commits/f17bdee8efd1dad3c480705104cae3cd478a7f98.

docs/06-concepts/04-authentication/04-providers/02-email/01-setup.md
marcelomendoncasoares: Addressed at https://github.com/serverpod/serverpod_docs/pull/707/commits/f17bdee8efd1dad3c480705104cae3cd478a7f98.

developerjamiu (approved):

PR #709, opened 2026-07-21
docs: Document dynamic model fields

marcelomendoncasoares:
## Summary

Models, endpoint parameters, and return values support `dynamic`, but nothing documented it. This adds a page for it and wires it into the places a reader would look first.

## Changes

- Add a "Dynamic fields" page under Models, covering when to reach for `dynamic` over a typed field, the model file syntax, nullability, the supported value types, endpoint use, JSON and JSONB persistence, `copyWith` behavior, and the tradeoffs.
- State that Serverpod carries runtime type metadata in the serialized value, and that the encoded form is an implementation detail.
- Cover project, module, and shared package model boundaries.
- Renumber the pages after it in the Models section.
- Link `dynamic` from the model type list, the endpoint parameter types, the database data representation section, and the model reference.

## Reference

- serverpod/serverpod#5013
- serverpod/serverpod#4996, serverpod/serverpod#3231
- Beta fixes: serverpod/serverpod#5146, serverpod/serverpod#5153, serverpod/serverpod#5157, serverpod/serverpod#5201


developerjamiu (approved):
LGTM

Issue #658, opened 2026-06-24
Add usage limits & quotas guide (plus input validation patterns)


developerjamiu:
## Demand

Kapa "Implementing Free Account Limits" cluster — 7 conversations / 3 users.

## Problem

Scopes/authorization work for tiers, but there is no guide on per-user usage quotas, and:

- No count-query vs counter-table trade-off guidance, or how to return a quota-exceeded error.
- No rate-limiting middleware example (the middleware page mentions rate limiting but shows none).
- Input validation patterns are not documented (response typing is).

## Scope

- A "Usage limits & quotas" guide: scopes for tiers + counter pattern + quota-exceeded exception + a rate-limit middleware example in `01-working-with-endpoints/04-middleware.md`.
- Input validation patterns (link to `04-exceptions.md`).

## Related

The endpoint-architecture half of this cluster is covered by #630.

Issue #626, opened 2026-06-17
Add how to guide on using Jaspr with Serverpod backend


Swiftaxe:
See https://docs.serverpod.dev/concepts/webserver/server-side-html#:~:text=%7D-,ALTERNATIVE,-If%20you%20prefer

PR #711, opened 2026-07-22
docs: Rework the Web server section for 4.0

developerjamiu:
Part 4 of the section-by-section Concepts rework (after #687, #693, and #696). This pass covers the Web server section: all 8 pages plus two pages that link into it. No page moves; retitles only (Request data, Server-side HTML, Single-page apps, Web server middleware H1).

Every claim was verified against the framework at 4.0.0-beta.0 and relic 1.2.0, and the key behaviors were confirmed on a live project: route matching, all header defaults via curl, the widget constructors, the middleware and session APIs, and the Jaspr sample.

## Corrections

- **Flutter web** rewritten: the page described tiered caching and a version-bump invalidation procedure that the framework removed (#4842); the real default is `no-cache, private` with ETag revalidation on every file. WASM is now documented as opt-in with `enableWasmHeaders`, matching what `serverpod create` scaffolds (#5257), and building goes through the generated `serverpod run flutter_build` task.
- **Overview**: `Response.created` does not exist in relic 1.2.0 (replaced with a table of the 13 real constructors and the generic `Response(201, ...)` form); route matching is by specificity, not registration order; the first-route example now compiles.
- **Non-compiling examples fixed across the section**: `await request.session` (it returns a `Future<Session>`), the module-grouping example's missing `handleCall`, the widget constructors (`ListWidget(widgets:)`, `JsonWidget(object:)`, `RedirectWidget(url:)`), `addMiddleware`'s required path argument, and a `session.db.find` call using the wrong API shape.
- **Cache guidance made consistent**: `publicImmutable` was paired with 5-minute lifetimes throughout the section, which contradicts what `immutable` means; plain serving now uses `public(maxAge:)`, and `publicImmutable` with a long lifetime is reserved for cache-busted assets. The helpers are documented as `cacheControlFactory` values with their verbatim headers, including the no-header default.

## New coverage

- Routes vs endpoints framing and a five-row decision table on the overview.
- A verified Jaspr sample on Server-side HTML: component, route, the mandatory `Jaspr.initializeApp()`, and the import aliasing it needs (jaspr 0.23).
- Health probe paths (`/livez`, `/readyz`, `/startupz`), the template dev loop (startup loading, reload and browser auto-refresh under `serverpod start`), `WidgetRoute` methods/caching behavior, body streaming with `request.read()` and `Body.fromDataStream`, and the note for projects created without the web server.
- The static-files, single-page-apps, and Flutter-web pages now link the Cloud CDN page, which already linked to them.

## Issues

Closes #366. Closes #365.

Advances #626 (Jaspr is now named across the section with a working sample; the full how-to guide remains open) and #368 (the broken example is fixed and `await request.session` plus the `ContextProperty` pattern are documented; the typed route-composition API ask stays with the framework). A rate-limiting example (#658) was considered and deferred. Part of the plan in #671.

### DIFF
--- a/docs/06-concepts/05-web-server/01-overview.md
+++ b/docs/06-concepts/05-web-server/01-overview.md
@@ -157,7 +157,7 @@ When you deploy, the web server ships with the rest of your project. On [Serverp
 
 - **[Routing](routing)** - Match requests to handlers by method and URL pattern
 - **[Request data](request-data)** - Access path parameters, query parameters, headers, and body
-- **[Middleware](./web-server-middleware)** - Intercept and transform requests and responses
+- **[Web server middleware](./web-server-middleware)** - Intercept and transform requests and responses
 - **[Static files](static-files)** - Serve static assets
 - **[Server-side HTML](server-side-html)** - Render HTML on the server with templates or Jaspr
 - **[Single-page apps](single-page-apps)** - Serve SPAs with client-side routing
--- a/docs/06-concepts/05-web-server/01-overview.md
+++ b/docs/06-concepts/05-web-server/01-overview.md
@@ -1,18 +1,20 @@
 ---
-description: Serverpod's built-in web server, built on the Relic framework, lets you serve web pages, REST APIs, webhooks, and static files alongside your application server.
+description: Serverpod's built-in web server serves REST APIs, webhooks, static files, server-rendered HTML, and Flutter web apps alongside the API server.
 ---
 
 # Overview
 
-In addition to the application server, Serverpod comes with a built-in web server. The web server allows you to access your database and business layer the same way you would from a method call from an app. This makes it simple to share data for applications that need both an app and traditional web pages. You can also use the web server to create webhooks or define custom REST APIs to communicate with third-party services.
+Serverpod comes with a built-in web server that runs beside the API server. It serves anything that speaks plain HTTP: REST APIs and webhooks, static files, server-rendered HTML with templates or [Jaspr](https://jaspr.site), and single-page apps including Flutter web. Web requests get the same [`Session`](../endpoints-and-apis/sessions) your endpoint methods receive, with full access to your database and business logic. The web server is built on the [Relic](https://github.com/serverpod/relic) framework, and its routing engine, middleware system, and typed headers are available directly.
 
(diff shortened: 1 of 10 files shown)


docs/06-concepts/05-web-server/01-overview.md
```diff
@@ -149,12 +149,16 @@ class UserRoute extends Route {
...
+## Going to production
+
+When you deploy, the web server ships with the rest of your project. On [Serverpod Cloud](../../deployments/deploy-to-serverpod-cloud), it is served through a CDN that honors the cache headers your routes set. See [Content delivery and caching](/cloud/concepts/cdn) for how the two interact.
+
 ## Next steps
 
 - **[Routing](routing)** - Match requests to handlers by method and URL pattern
-- **[Request Data](request-data)** - Access path parameters, query parameters, headers, and body
+- **[Request data](request-data)** - Access path parameters, query parameters, headers, and body
 - **[Middleware](./web-server-middleware)** - Intercept and transform requests and responses
```
Zfinix: Link text still says "Middleware" but this PR retitles the page to "Web server middleware". Match the link text to the page title, especially since disambiguating from endpoint middleware was the reason for the retitle.

docs/06-concepts/05-web-server/06-server-side-html.md
```diff
@@ -109,19 +114,69 @@ class DataRoute extends WidgetRoute {
...
+jaspr.Jaspr.initializeApp();
+
+pod.webServer.addRoute(JasprHelloRoute(), '/jaspr');
+```
+
+:::warning
+The `initializeApp()` call is required. Without it, the first render kills the whole server process instead of failing the single request.
 :::
 
+Two details make this example work. Both `package:jaspr/server.dart` and `package:jaspr/dom.dart` are imported under one alias, because `server.dart` exports names that collide with the web server's `Request` and `Response`, and the HTML components (`div`, `h1`, `p`) live in `dom.dart`. And `renderComponent` returns a result record whose `body` is bytes, which `utf8.decode` turns into the HTML string. The example was verified against jaspr 0.23.
```
Zfinix: `The example was verified against jaspr 0.23.` reads as an author verification note rather than reader guidance. Cut it, or fold the version into the `dart pub add jaspr` step so it lands where the reader acts on it.

Zfinix (approved):
LGTM 🚀 Approved once the two inline nits are resolved.

PR #719, opened 2026-07-27
docs: Rework the Scheduling section for 4.0

developerjamiu:
## Summary

Section-by-section rework of the Scheduling (future calls) concept pages for Serverpod 4.0, following the same pattern as the Data & the database (#696) and Web server (#711) passes. Adds a section overview and brings every page current with the 4.0 API, verified against the framework source.

## Changes

- **New Overview page.** The section now opens with a landing page: what a future call is, the three scheduling shapes, the at-least-once execution guarantee, and the run-mode facts. Inbound links retarget here.
- **Future calls** (was Setup): explains the generated `.example` accessor, uses the standard `serverpod start` / `serverpod generate` framing, and corrects the UTC behavior of `callAtTime`.
- **Recurring tasks** (rewritten): documents the native `callRecurring().cron()` / `.every()` API, which was previously undocumented. Cron schedules are noted as UTC. The manual self-reschedule pattern is kept as a short fallback for dynamically computed intervals.
- **Inheritance** (reframed): centers the real use case of a module exposing future calls through an abstract base, and fixes an example that claimed to share behavior through `super`.
- **Configuration**: options table, corrected concurrency semantics (`0` also means unlimited), and a corrected note that `--role maintenance` executes due calls rather than only checking for broken ones.
- **Legacy**: states the `@Deprecated` status of the string-based scheduling methods precisely.

## Notes

- Execution is at-least-once (not exactly-once); the previous wording overstated the guarantee.
- All framework claims verified against 4.0.0-beta.0.


Zfinix (commented):
The two renamed URLs here need redirects too, same list as the testing PR.

Zfinix (changes requested):

docs/06-concepts/06-scheduling/03-recurring-tasks.md
```diff
@@ -0,0 +1,97 @@
...
+    .cleanUp();
+```
+
+A `start` in the past runs the call at the next scan, then continues on the interval from there. To pin a task to a wall-clock time such as 02:00 every day, use a cron expression instead.
+
+If the server is down when a run is due, that run happens once when the server is back, then the schedule jumps ahead to the next interval boundary. The intervals missed while the server was down do not pile up and fire all at once.
+
+## Repeat on a cron schedule
+
+Use `cron` for calendar-based schedules that a fixed interval cannot express, such as "every day at 02:00" or "every Monday". A cron expression is a short text format of five fields, `minute hour day-of-month month day-of-week`, where `*` means "every".
```
Zfinix: Cron.parse takes 5 or 6 fields, the 6th being seconds. As written, someone with a 6 field expression will think it is invalid.

docs/06-concepts/06-scheduling/05-configuration.md
```diff
@@ -0,0 +1,95 @@
...
+- **Unset (the default):** the check runs only when the table holds fewer than 1000 future calls, so a large backlog does not slow startup.
+- **`true`:** the check always runs, regardless of how many calls are scheduled.
+- **`false`:** the check never runs.
+
+```yaml
+futureCall:
+  checkBrokenCalls: true
+```
+
+:::note
```
Zfinix, line 78: This note is doing three jobs, the role, the command and a production warning. The part about not running it against a production database should be its own warning.

docs/06-concepts/06-scheduling/05-configuration.md
```diff
@@ -0,0 +1,95 @@
...
+The `maintenance` role also runs this check. Started in that role, the server checks for broken calls, runs every currently-due future call once, and then exits. This is how future calls get processed in environments that do not run a persistent server, such as serverless hosting. See [server roles](../server-fundamentals/running-your-server#choose-a-server-role).
+
+```bash
+$ dart run bin/main.dart --role maintenance
```
Zfinix: $ prefix here. None of the other bash blocks in the section have it.

docs/06-concepts/06-scheduling/02-future-calls.md
```diff
@@ -0,0 +1,128 @@
...
+    args,
+    Protocol(),
+    Endpoints(),
+  );
+
+  await pod.start();
+}
+```
+
+:::note
```
Zfinix, line 54: Two separate points in one note, worth splitting.

docs/06-concepts/06-scheduling/02-future-calls.md
```diff
@@ -0,0 +1,128 @@
+---
+description: Define a future call, schedule it after a delay or at a specific time with the type-safe API, and cancel scheduled calls by identifier.
```
Zfinix: Description should open with the noun, not the verb. Same on recurring-tasks, inheritance and configuration.

docs/06-concepts/06-scheduling/01-overview.md
```diff
@@ -0,0 +1,35 @@
...
+
+If the method itself throws an exception, Serverpod logs the error and does not retry the call. You are responsible for retrying failed work: schedule a new call if the work needs to happen. Recurring calls are the exception, since the next run is always scheduled regardless of whether the current one succeeds.
+
+## Where future calls run
+
+Which servers run future calls depends on their role, the mode a server is started in. Servers in the default `monolith` role run them, and future calls need a database. A server started in the `serverless` role does not run them. If you host in a serverless environment, run a separate scheduled process in the `maintenance` role to execute due calls. See [server roles](../server-fundamentals/running-your-server#choose-a-server-role) and [hosting elsewhere](../../deployments/custom-hosting/hosting-elsewhere).
+
+## In this section
```
Zfinix: Operations calls this What each page covers, and none of the pages here have a Related block while Operations and Testing have one on every page. Worth settling on one shape for the series.

docs/06-concepts/06-scheduling/04-inheritance.md
```diff
@@ -0,0 +1,91 @@
+---
+description: FutureCall inheritance shares methods between classes, exposes a module’s future calls through an abstract base, and allows overrides.
```
Zfinix: Nit: Curly apostrophe in "module's", the rest of the docs use a straight one.

Zfinix (approved):
LGTM 🚀 approved when the apostrophe is fixed.

PR #720, opened 2026-07-27
docs: Rework the Operations section for 4.0

developerjamiu:
## Summary

Section-by-section rework of the Operations concept pages for Serverpod 4.0, following the Data & the database (#696), Web server (#711), and Scheduling (#719) passes. The section gains an entry point, loses the two topics that did not belong to it, and every remaining page was checked against the framework source.

## Changes

- **New Overview page.** The section now opens with what Operations covers and links the pages under it.
- **Caching moves to Endpoints & APIs**, next to Sessions and Server events, since it is a `Session` facility rather than a run-in-production concern. The page now documents invalidation, cache groups, the entry limit and eviction order, and the four caches rather than three.
- **Experimental features is replaced by Exception monitoring**, named for its actual subject. Shutdown tasks move to Running your server, the diagnostic-handler test example moves to Advanced testing examples, and the CLI flag table (whose only row enabled nothing in 4.0) is dropped.
- **Logging**: separates the four kinds of record, tables the settings with their environment variables and real defaults, and documents runtime log settings and the conditions under which log purging is disabled.
- **Health checks**: separates the HTTP probes from the metrics collection, which share a name but are unrelated APIs, and corrects the response format and metric cadence.
- **Security and TLS**: leads with the common case (Serverpod Cloud terminates TLS for you, a proxy usually does otherwise) before the case where the server does it itself.

## Notes

- Two code samples in the section did not compile as written and are fixed: the `CacheMissHandler` example returned a nullable value from a non-nullable method, and the `ServerHealthMetric` example omitted the required `granularity` argument.
- `/startupz` no longer claims to verify migrations or to govern traffic routing; readiness governs routing.
- The global cache falls back to an in-memory, per-instance cache whenever Redis is not configured, in every run mode. This was previously documented as a development-only fallback.
- Cloud logs and the Redis guide now link to their framework counterparts.


Zfinix (commented):
Caching moving to endpoints-and-apis and experimental-features going away both change URLs, they need redirects too.

docs/06-concepts/07-operations/02-logging.md
```diff
@@ -21,86 +32,86 @@ session.log(
...
+Session logging is configured under `sessionLogs:` in your config file for the run mode, or through environment variables. Environment variables win over the config file, key by key.
 
-  - `persistentEnabled` is set to `true`, meaning logs are stored in the database.
-  - `consoleEnabled` defaults to `true` in the `development` run mode and `false` in the other run modes, so local runs print to the console while deployed runs rely on the database logs unless console logging is explicitly enabled.
+| Setting | Environment variable | Default |
+| --- | --- | --- |
+| `persistentEnabled` | `SERVERPOD_SESSION_PERSISTENT_LOG_ENABLED` | `true` when a database is configured |
+| `consoleEnabled` | `SERVERPOD_SESSION_CONSOLE_LOG_ENABLED` | `true` in `development` or when there is no database, otherwise `false` |
+| `consoleLogFormat` | `SERVERPOD_SESSION_CONSOLE_LOG_FORMAT` | `text` in `development`, otherwise `json` |
+| `cleanupInterval` | `SERVERPOD_SESSION_LOG_CLEANUP_INTERVAL` | `24h`, but see [Purge old records](#purge-old-records) |
```
Zfinix: Nit: these three defaults only apply when sessionLogs is missing entirely, but generated projects always ship the block, so the table is showing the case that basically never happens. put the real behaviour in the cell

docs/06-concepts/07-operations/02-logging.md
```diff
@@ -21,86 +32,86 @@ session.log(
...
+  cleanupInterval: 6h       # Purge every 6 hours
+  retentionPeriod: 30d      # Keep 30 days
+  retentionCount: 5000      # Keep at most 5,000 sessions
+```
 
-  - `persistentEnabled` is set to `false` since persistent logging requires a database.
-  - `consoleEnabled` is set to `true`, meaning logs are printed to the console by default.
+Durations use the same format as [model default values](../data-and-the-database/models#supported-default-values), such as `30d`, `6h`, or `1d 2h 30min`.
 
 :::warning
```
Zfinix, line 87: Two unrelated warnings in one block, the StateError one and the SQLite one. Should be two.

docs/06-concepts/07-operations/02-logging.md
Zfinix: These three defaults only apply when sessionLogs is missing entirely, but generated projects always ship the block, so the table shows the case that basically never happens. Put the real behaviour in the cell.

docs/06-concepts/07-operations/05-exception-monitoring.md
```diff
@@ -0,0 +1,72 @@
...
+      );
+    }
+    return 'success';
+  }
+}
+```
+
+The `ExceptionEvent` class is the built-in event for a thrown exception and a stack trace. Passing the session attaches its context to the event, so the handler knows which call the exception came from. This works anywhere you have a session, including endpoint methods, web calls, and future calls.
+
+## Experimental APIs in general
```
Zfinix: This reads like leftover from the deleted experimental-features page, and it is a second subject on a page named for one thing. Could be a lead sentence, or live under Server fundamentals.

docs/06-concepts/02-endpoints-and-apis/06-caching.md
```diff
@@ -0,0 +1,153 @@
+---
+description: Cache objects in server memory or in Redis through session.caches to avoid repeating expensive database work, with lifetimes, groups, and explicit invalidation.
```
Zfinix: Description opens with a verb, same on exception-monitoring. Should lead with the noun.

docs/06-concepts/07-operations/03-health-checks.md
```diff
@@ -1,118 +1,82 @@
 ---
-description: Health checks in Serverpod expose Kubernetes-style HTTP endpoints (/livez, /readyz, /startupz) for liveness, readiness, and startup probing, with support for custom health indicators and metrics.
+description: Serverpod answers liveness, readiness, and startup probes at /livez, /readyz, and /startupz, supports custom health indicators, and collects health metrics about itself.
 ---
 
 # Health checks
 
-Serverpod provides a complete health check system that allows you to monitor the health of your server and your dependencies through Kubernetes-style HTTP endpoints (`/livez`, `/readyz`, `/startupz`) - each with a specific purpose that helps orchestrators (like Kubernetes) make informed decisions about container lifecycle and traffic routing.
+When your server runs behind a host that can restart it or route traffic away from it, that host needs a way to ask how the server is doing. Serverpod answers three such questions over HTTP, at URLs that match what container platforms like Kubernetes expect.
```
Zfinix, line 8: The page says up front that it is two unrelated things, then needs a note at the bottom to stop people mixing them up. Probably two pages. Your call, not blocking.

developerjamiu: Deliberate call when we set the section structure. Metrics on its own would be a short page few readers reach, and the collision people actually hit is `healthConfig`/`HealthIndicator` against `healthCheckHandler`/`ServerHealthMetric`, which is easier to tell apart side by side than across two pages.

Zfinix (approved):
LGTM 🚀

PR #721, opened 2026-07-27
docs: Rework the Testing section for 4.0

developerjamiu:
## Summary

Section-by-section rework of the Testing concept pages for Serverpod 4.0, following Data & the database (#696), Web server (#711), Scheduling (#719), and Operations (#720). The section keeps its linear shape but splits reference material out of the walkthrough, and every claim was checked against the framework source.

Stacked on #720, so it targets that branch and shows only its own changes. GitHub retargets it to `main` once #720 merges.

## Changes

- **Get started** drops from 229 lines to 66. The collapsed "Serverpod 2.1 or earlier" block, which duplicated the upgrade guide near-verbatim, becomes a pointer to it. The page now opens by saying what the test tools do before the setup steps.
- **Writing tests** (was "The basics") keeps the pieces every test uses: the session builder, simulated authentication, seeding, and the exceptions the tools throw.
- **Configuration** is a new page holding every `withServerpod` option, so the reference table no longer interrupts a page being read top to bottom.
- **Advanced examples** gains future-call testing and loses a section whose advice did nothing (see below).
- **Best practises** becomes **Best practices**, matching the spelling used everywhere else in the docs.

## Corrections

- The Docker prerequisite is gone. A 4.0 project sets `database.dataPath` in `config/test.yaml`, so `dart test` starts its own PostgreSQL and the compose services are never contacted.
- "Optimizing the number of database connections" claimed the Serverpod instance is created lazily by `sessionBuilder.build()`, and recommended moving `build()` into `setUpAll` to reduce connections. The instance is constructed when `withServerpod` is called and the connection opens in `setUpAll` regardless, so the workaround changed nothing.
- The configuration table was missing `serverDirectory`, and presented `applyMigrations`, `rollbackDatabase`, and `runtimeParametersBuilder` as always available. Without a database they are not generated at all, so passing one is a compile error.
- The `flushEventQueue` example passed a built `Session` where the generated API takes the session builder.
- `copyWith` defaults were documented as fixed values; both parameters inherit from the builder they are called on.
- The generated file does not re-export Serverpod's own types, so pages that use `Session`, `Scope`, or `Constant` now say where they come from.

## Notes

- Two renamed URLs (`/concepts/testing/the-basics`, `/concepts/testing/best-practises`) need redirects at the GA cutover; recorded with the rest of the redirect work.
- Three framework issues surfaced while verifying and are worth filing separately: the `serverpod_test` doc comment for `flushEventQueue` has the same `session`/`sessionBuilder` error corrected here, the start-timeout message tells users to run `docker compose up`, and `working-with-endpoints.md` has a pre-existing broken link to this section.


docs/06-concepts/08-testing/03-configuration.md
```diff
@@ -0,0 +1,151 @@
+---
+description: Every option withServerpod accepts, including run mode, database rollback behavior, server output, timeouts, and test tags.
+---
+
+# Configuration
+
+The `withServerpod` helper takes optional named arguments that control how the test server starts and how tests are isolated from each other.
+
+Some of the values below, such as `ServerpodRunMode` and `Constant`, come from `package:serverpod/serverpod.dart` rather than the generated test tools, so import it when you use them.
```
Zfinix: serverDirectory takes a Directory from dart:io, worth naming here since this line is already telling people where the types come from.

docs/06-concepts/08-testing/01-get-started.md
```diff
@@ -1,229 +1,63 @@
...
+Projects created with Serverpod 2.1 or earlier need a few one-time setup steps first. See [Upgrade to 2.2](../../upgrading/archive/upgrade-to-two-point-two).
+:::
 
-Go to the server directory and generate the test tools:
+The test tools are generated along with the rest of your code. With `serverpod start` running, saving a file regenerates them. Outside a session, run `serverpod generate` from your server directory.
 
-```bash
-serverpod generate
-```
+Either way you get `test/integration/test_tools/serverpod_test_tools.dart`. The `test/integration` folder keeps these apart from your unit tests, which makes a project easier to navigate. Running the two groups separately is a different mechanism: `withServerpod` tags its tests `integration`, and you filter on that tag. To generate the file somewhere else, change `server_test_tools_path` in `config/generator.yaml`.
```
Zfinix: The sentence about tagging and filtering is a forward reference in the middle of setup, and the detail already lives on advanced examples. Would cut it and let the paragraph run from the file straight to changing the path.

docs/06-concepts/08-testing/05-best-practices.md
```diff
@@ -7,27 +7,28 @@ toc_max_heading_level: 2
...
+// The generated file carries the test helpers and your endpoints.
+import 'test_tools/serverpod_test_tools.dart'; ✅
 ```
 
+Serverpod's own types are a separate matter. The generated file does not re-export them, so a test that uses `Session`, `Scope`, `Constant`, `ServerpodRunMode`, `ExperimentalFeatures`, or a Serverpod exception type needs `package:serverpod/serverpod.dart` as well. Your own models come from your project's generated protocol.
+
 ## Database clean up
 
-Unless configured otherwise, by default `withServerpod` does all database operations inside a transaction that is rolled back after each `test` (see [the configuration options](./the-basics#rollback-database-configuration) for more info on this behavior).
+Unless configured otherwise, by default `withServerpod` does all database operations inside a transaction that is rolled back after each `test` (see [the configuration options](./configuration#rollbackdatabase) for more info on this behavior).
```
Zfinix: `./configuration#...` here, the rest of the section uses the bare `configuration#...` form.

docs/06-concepts/08-testing/01-get-started.md
```diff
@@ -1,229 +1,63 @@
 ---
 sidebar_label: Get started
-description: Integration testing in Serverpod uses the withServerpod helper to call endpoints directly, seed the database, and simulate authenticated sessions without running a separate server.
+description: Test Serverpod endpoints by calling them like ordinary Dart functions, with a real server and a real database that resets between tests.
```
Zfinix: Nit: Description opens with a verb, could lead with the noun.

Zfinix (commented):
Two renamed URLs here, worth getting the redirects in with the rest.

Zfinix (approved):
LGTM 🚀

PR #724, opened 2026-07-30
docs: Clarify Google sign-in setup for web and Android

developerjamiu:
Fixes friction found by walking the Google sign-in setup guide end to end against a fresh 4.0.0-beta.1 project, testing on Android (emulator) and web.

## Setup page

- **Web:** the same-origin constraint and the "`flutter run -d chrome` doesn't work" caveat move into a warning placed before the build command, since setting up around the dev server is the natural mistake and the page only warned about it afterwards. The build command gains `--base-href /app/` (without it the app loads blank under `/app`; `FlutterRoute` serves files as-is, and the project template's own `flutter_build` script includes the flag) and a "replace `my_project_server`" sentence matching the page's placeholder pattern.
- **Web:** new warning that the app served at `/app` is a compiled snapshot, so `main.dart` changes need a rebuild. Testing showed a stale build is an easy way to hit `redirect_uri_mismatch` after correcting a redirect URI.
- **Android:** the download-and-rename `google-services.json` steps are removed. The Auth Platform download is a client-secret JSON, not a `google-services.json`, and a plain Flutter project has no `com.google.gms.google-services` Gradle plugin to read one, so the step cannot supply the server client ID. Passing `serverClientId` (the Web client's ID) is now the primary path, shown in the initialize snippet's mobile branch; `google-services.json` remains as a note for Firebase-based projects. Verified against the failure `GoogleSignInException(clientConfigurationError, serverClientId must be provided on Android)` and confirmed working after the change.
- **Android:** warning that the emulator cannot reach `localhost`, linking the new troubleshooting entry.

## Troubleshooting page

- New `redirect_uri_mismatch` cause: stale web build.
- The "Missing web client entry in google-services.json" entry is retitled to the actual error message ("serverClientId must be provided") and rewritten with the real cause and the programmatic fix; the Firebase re-download guidance is kept for projects that use the Gradle plugin.
- New entry: endpoint calls fail on Android with connection refused (template `assets/config.json` pins `localhost`, which overrides the framework's `10.0.2.2` emulator handling; resolution via `--dart-define=SERVER_URL=...`).
- Setup checklist updated to match the Android changes.

Follow-up (not in this PR): the template shipping `assets/config.json` with `localhost` affects every scaffolded project on Android and is being raised as a framework issue; the endpoints page's "localhost adapts to the platform" wording should be revisited once that lands.

### DIFF
--- a/docs/06-concepts/04-authentication/04-providers/03-google/04-troubleshooting.md
+++ b/docs/06-concepts/04-authentication/04-providers/03-google/04-troubleshooting.md
@@ -201,6 +201,18 @@ On the Android emulator, `10.0.2.2` maps to the host machine. On a physical devi
 
 **Resolution:** In the running `serverpod start` terminal, press **M** to create the migration, then **A** to apply it.
 
+## Google sign-in button does not appear
+
+**Problem:** `SignInWidget` renders, but the Google button is missing.
+
+**Cause:** `SignInWidget` shows the Google button when the client has a registered `GoogleIdpEndpoint` and the Google sign-in service is initialized. The common misses:
+
+- The app was hot reloaded after adding `initializeGoogleSignIn` to `main.dart`. Hot reload does not re-run `main()`, so the service is never initialized.
+- `GoogleIdpEndpoint` is missing on the server, or the client was not regenerated after adding it.
+- On web, `initializeGoogleSignIn` was called without `clientId` and `redirectUri`. The widget renders nothing without them.
+
+**Resolution:** Hot restart the app: press **R** in the `serverpod start` terminal, or rerun `flutter run`. If the button is still missing, confirm `GoogleIdpEndpoint` exists on the server and run `serverpod generate`, and on web confirm `initializeGoogleSignIn` receives `clientId` and `redirectUri` per [Web setup](./setup#web).
+
 ## Lightweight sign-in (One Tap) not appearing
 
 **Problem:** You enabled `attemptLightweightSignIn: true` but the One Tap prompt never appears on Web, or the silent sign-in doesn't trigger on mobile.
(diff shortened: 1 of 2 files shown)


docs/06-concepts/04-authentication/04-providers/03-google/04-troubleshooting.md
```diff
@@ -152,13 +153,37 @@ Every line of the JSON must be indented by at least one level more than `googleC
...
+
+You can also supply it at build time with `--dart-define=GOOGLE_SERVER_CLIENT_ID=...`; see [Configuring Client IDs on the App](./customizations#configuring-client-ids-on-the-app).
+
+For Firebase-based projects using the Gradle plugin, make sure a Web application OAuth client exists in the same Google Cloud project and re-download `google-services.json` so it includes the web client entry.
+
+## Endpoint calls fail on Android with connection refused
+
+**Problem:** Sign-in completes at Google, but the app then fails with `ServerpodClientException: ... Connection refused ... uri=http://localhost:8080/...`.
+
+**Cause:** On Android, `localhost` is the emulator or device itself, not the machine running your server. The project template's `assets/config.json` sets `apiUrl` to `http://localhost:8080`, and that value takes precedence over the framework's platform-aware default (see [server URL resolution](../../../endpoints-and-apis)).
```
Zfinix, line 178: Does `../../../endpoints-and-apis` resolve for you? It points at the section folder, which has no index page. Worth checking.

developerjamiu: It resolves: `01-working-with-endpoints.md` declares `slug: /concepts/endpoints-and-apis`, so the page owns the section URL (the category has `link: null`). The build's link check fails on the `/working-with-endpoints` form and passes on this one.

Zfinix (approved):
LGTM 🚀

PR #665, opened 2026-07-01
docs: Add macOS Keychain Sharing guide for authentication (#4512)

abdulawalarif:
## Summary

- Adds a new guide, [Set up authentication on macOS](docs/06-concepts/11-authentication/08-macos-authentication.md), documenting the Keychain Sharing entitlement required for sign-in sessions to persist on macOS.
- Covers the default fix (empty `keychain-access-groups` in entitlements files), verification steps, Xcode setup, and a local-development alternative using the login keychain.
- Adds troubleshooting for the `PlatformException(-34018, A required entitlement isn't present, ...)` error reported in serverpod/serverpod#4512.
- Links to the new page from Get started, Setup, Authentication basics, and the Facebook provider macOS section.

Closes serverpod/serverpod#4512 (docs portion).

## Test plan

- [ ] Run `npm run build` locally and confirm the site builds without broken links
- [ ] Confirm the new page appears under Authentication in the sidebar
- [ ] Verify cross-links resolve: `./macos-authentication` from get-started, setup, basics, and Facebook setup
- [ ] Review the entitlements XML snippet and troubleshooting steps against a macOS Flutter project

PR #728, opened 2026-07-31
docs: Rework the Authentication core and platform pages for 4.0

developerjamiu:
First of three PRs reworking the Authentication section for 4.0. This one covers the core and platform pages. Providers and the legacy section follow separately.

Every framework claim here was verified against `serverpod/serverpod` at 4.0.0-beta.1.

## Structure

`01-get-started.md` and `01-setup.md` both carried the `01-` prefix, so sidebar order depended on an alphabetical tiebreak. The pages are renumbered so each number is used once. Slugs strip numeric prefixes, so **no URLs change**.

The macOS authentication page from #665 is now `09-`, keeping the order that PR intended (profile photos, then macOS, then custom overrides). Its three inbound notes are preserved in the rewritten pages.

## Correctness

The larger fixes:

- **Custom overrides, authentication schemes.** The page said Serverpod defaults to the `basic` scheme and that unrecognized schemes reach your handler. Neither is true. The client sends whatever `authHeaderValue` returns, the built-in module uses `bearer`, and the server accepts only `Bearer`, `Basic`, and `Digest`, rejecting anything else with a 400 unless `validateHeaders` is `false`.
- **The custom-overrides example did not work.** It wrapped the key `'valid'` with `wrapAsBasicAuthHeaderValue`, but `Basic` requires the decoded value to be `username:password` with both parts non-empty, so the request was rejected with a 400 before reaching the handler. Switched to `wrapAsBearerAuthHeaderValue`, which matches what the built-in module does.
- **`googleClientSecret` showed a service account JSON.** The framework parses the OAuth client JSON with a `web` section and throws at startup otherwise.
- **The web callback section omitted Microsoft** and custom OAuth2 providers, and did not mention that the server-served `FlutterWebAuth2CallbackRoute` only works when the app is on the same origin.
- `AuthServices.getTokenManager<T>()` is static, not an instance member.
- The `userProfile` extension throws `UserProfileNotFoundException` rather than returning `null` when a signed-in user has no profile.
- `JwtConfigFromPasswords` supports HMAC SHA-512 only.
- `defaultUserImageGenerator` draws a square, not a circle.
- Profile photos are re-imported on any sign-in where the user has no image, not only the first.
- Both token-manager pages claimed the client redirects to a login page on expiry. It signs the user out; navigation is the app's job.
- The server-side sessions page advertised immediate scope updates on existing sessions. Sessions snapshot their scopes at creation and nothing updates them.
- Client-side exception types on the profile photos page, the `AuthUsers.delete` named parameter, the module path `client.modules.serverpod_auth_core.status`, and the `AuthenticationInfo` contract (`userIdentifier` is a `String`, `authId` is required).
- Version pins moved from `4.0.0-beta.0` to `4.0.0-beta.1`.

## Currency

The provider list named four of the nine shipped providers. It now names all nine and marks the two experimental ones. `FlutterWebAuth2CallbackRoute` is documented as an alternative to hand-creating `auth.html`. The shared button styling API (`SignInWidget.buttonStyle`) is documented, replacing per-provider enums that no longer exist.

## Readability

Jargon that was never explained now is, at first use: pepper, identity provider, access versus refresh tokens, and the difference between the `Session` request object and an auth session. The token managers page gained a comparison table so the choice between JWT and server-side sessions is answerable on the page that presents it.

Duplicated content was given a single owner and a pointer: the `SignInWidget` example, the default user image configuration, and the do-not-navigate warning. Every page ends with a Related block.

## Verification

Two adversarial review passes ran over this change: one over the rework, one over the fixes made in response to the first. The second pass caught nine problems in the first pass's own corrections, including the broken `basic` example above.

`npm run build` passes.


docs/06-concepts/04-authentication/02-setup.md
```diff
@@ -64,7 +62,10 @@ void run(List<String> args) async {
   await pod.start();
 }
 ```
-Then extend the abstract endpoint for refreshing JWT tokens to expose it on the server. Create the file anywhere under your server's `lib/` directory (for example, `<project>_server/lib/src/endpoints/`); the generator picks it up:
+
+The pepper in the example is a secret string mixed into values before they are hashed. The [storing secrets](#storing-secrets) section explains peppers and where to keep them.
```
Zfinix: Pepper is defined here and again at line 138 in nearly the same words. This one already links to Storing secrets, so it can drop the definition and keep the pointer.

docs/06-concepts/04-authentication/04-working-with-users.md
```diff
@@ -8,25 +8,25 @@ The authentication module provides convenient ways to work with your authenticat
...
 ## Authenticated users
 
-All authenticated users have an authentication identifier, that uniquely identifies them across the server. This can be retrieved from the `session` object as a `String` through the `userIdentifier` property or as a `UuidValue` from the `authUserId` extension provided by the authentication module.
+All authenticated users have an authentication identifier that uniquely identifies them across the server. This can be retrieved from the `session` object as a `String` through the `userIdentifier` property or as a `UuidValue` from the `authUserId` extension provided by the authentication module.
 
 ```dart
 var userIdString = session.authenticated?.userIdentifier;
 // requires `import 'package:serverpod_auth_idp_server/core.dart';`
-var userIdUuidValue = session.authenticated?.authUserId;
+var userIdUuidValue = session.authenticated!.authUserId;
```
Zfinix: Nit: `!.` here, but `?.` on the line above and again at line 256 for the same call. Worth picking one.

docs/06-concepts/04-authentication/02-setup.md
```diff
@@ -248,14 +254,14 @@ macOS apps need a Keychain Sharing entitlement before authentication sessions ca
...
 
 You create one `auth.html` and share it across every identity provider that needs it.
 
-In your Flutter project's `web/` folder, add a file named `auth.html` with this content:
+You have two ways to deliver it. If Serverpod also serves your Flutter web app, register the `FlutterWebAuth2CallbackRoute` from `serverpod_auth_idp_server` on your web server and point the provider at that route. The page hands the result back with `postMessage` aimed at its own origin, so the browser only delivers it when your app is served from the same origin (same scheme, host, and port). Otherwise, host the file yourself: in your Flutter project's `web/` folder, add a file named `auth.html` with this content, which is identical to what the route serves:
```
Zfinix: Both options, the same-origin constraint and the code lead-in sit in one paragraph, and the closing colon only introduces the second option. Splitting reads better:

```suggestion
You have two ways to deliver it.

If Serverpod serves your Flutter web app, register the `FlutterWebAuth2CallbackRoute` from `serverpod_auth_idp_server` on your web server and point the provider at that route. The page posts the result back to its own origin, so the browser only delivers it when your app is served from that same origin (same scheme, host, and port).

Otherwise, host the file yourself. In your Flutter project's `web/` folder, add a file named `auth.html` with this content, which is identical to what the route serves:
```

Zfinix (approved):
Approved when changes are applied 🚀

docs/06-concepts/04-authentication/02-setup.md
```diff
@@ -151,12 +156,12 @@ export SERVERPOD_PASSWORD_serverSideSessionKeyHashPepper='your-session-pepper-he
...
 export SERVERPOD_PASSWORD_jwtHmacSha512PrivateKey='your-private-key-here'
 export SERVERPOD_PASSWORD_emailSecretHashPepper='your-email-pepper-here'
-export SERVERPOD_PASSWORD_googleClientSecret='{"type":"service_account",...}'
+export SERVERPOD_PASSWORD_googleClientSecret='{"web":{"client_id":"...","client_secret":"...","redirect_uris":["..."]}}'
 # ... other secrets
 ```
 
 :::info
-When using the `config/passwords.yaml` file or environment variables, you can use a convenience version of token manager and identity provider builders that already load secrets using `pod.getPassword()` while still allowing you to pass additional configuration options.
+Builders that need secrets have a `FromPasswords` variant that reads them from well-known key names, so you do not need to call `pod.getPassword()` yourself. Any other configuration options are still passed as parameters. The examples below are a few of them:
```
Zfinix: "The examples below are a few of them" is vague about what "them" is. "For example:" says the same thing.

PR #729, opened 2026-08-05
docs: Correct the Authentication provider pages for 4.0

developerjamiu:
Second of three PRs reworking the Authentication section for 4.0, after #728. This one is a correctness-only pass over the provider pages: no files move, no pages merge. The structural normalization of the provider groups follows in a separate PR so each can be reviewed for one thing.

Every framework claim was verified against `serverpod/serverpod` at 4.0.0-beta.1.

## The systemic fix

Six provider UI pages documented per-provider button enum families (`GSIButtonTheme`, `FacebookButtonText`, `MicrosoftButtonType`, and friends) that do not exist at 4.0. Every styling snippet on those pages failed to compile. They now use the real shared API (`SignInButtonSize`/`Shape`/`TextVariant` plus the per-brand `*ButtonStyle` enums), with the actual constructor defaults shown, and a note that brand color presets only apply outside `SignInWidget`.

## Documented flows that did not work

- The custom OAuth2 client flows awaited the endpoint and discarded the returned `AuthSuccess`, so the app was never signed in. They now register it with `client.auth.updateSignedInUser`. The endpoint snippets also missed the abstract `hasAccount` override and, on the walkthrough page, the `IdentityProvider` contract, so none of them compiled.
- The Firebase sign-in gate never listened to its controller, so the UI never rebuilt after a successful login.
- Apple's Android manifest registered a different package's callback activity, so sign-in could never return to the app. The correct `SignInWithAppleCallback` block (from the `sign_in_with_apple` package) is now used in both places it appears.

## Pressing M applies the migration too

The `serverpod start` TUI creates and applies a migration in one step when you press **M**; **A** retries a failed apply. Fifteen pages in this section said "press M to create the migration, then A to apply it", including one line in the email setup page from #707. All are corrected, and the setup page documents the retry role of **A**. The same wording exists in two pages outside this section (`server-fundamentals/modules` and `database/tables`); those are left for a follow-up so this PR stays in one section.

## The Flutter app is served at /, not /app

Default projects (`website=false`, the default in both the TUI and non-interactive create) mount `FlutterRoute` at `/`; only website-enabled projects use `/app`. The GitHub local web-test command (`--base-href /app/`) would have produced a broken app on a default project, and Google troubleshooting hardcoded the `/app` path. Both now describe the default with a note for the website variant.

## Other corrections worth naming

- Apple: `webAuthenticationCallbackRoutePath` default is `/auth/apple/callback`; `serviceIdentifier`/`redirectUri` are required on every platform (initialization throws without them); Serverpod already revokes tokens on Apple's revocation notification, so the manual-invalidation guidance is gone; authorization codes expire in five minutes, not ten; `onAfterAppleAccountCreated` exists and is now documented with an example.
- Google: the "Enable People API" setup section is removed (only the legacy module ever called it; 4.0 uses the userinfo endpoint); the fictional no-`redirectUri` web mode is gone (web always uses the redirect flow); `attemptLightweightSignIn` has no effect on web; the default account validation requires `verifiedEmail` plus both name fields; `String.fromEnvironment` calls are now `const` so `--dart-define` values are actually read.
- GitHub: Serverpod does not store or refresh the GitHub access token, so the pages no longer claim it does; `redirectUri` is a `String`, not a `Uri`; the private-email troubleshooting reflects the `/user/emails` fallback; scope changes from callbacks apply on the next sign-in.
- Firebase: the beta.1 default accepts unverified emails, with `FirebaseIdpConfig.requireVerifiedEmail` as the opt-in and `FirebaseEmailNotVerifiedException` reaching the app; only `project_id` is used from the service account JSON; `findUserByFirebaseUserId` is static; the "invalid signature" troubleshooting names the real causes (verification is against Google's public certificates, not the service key).
- Facebook: `FacebookAccountDetails` exposes `fullName`/`firstName`/`lastName`, not `name`; the client token is used by the mobile SDKs, not web.
- Microsoft: `MicrosoftAccountDetails.imageBytes` (`Uint8List?`), not `image` (`Uri?`); the `MICROSOFT_TENANT` dart-define is removed since nothing reads it; the profile photo is fetched after validation, controlled by `fetchProfilePhoto`.
- Email: the custom validator example throws `InvalidEmailException` so the widget can show the message; the development bypass code is digits-only, matching what the verification input accepts; `deleteEmailAccountRequestById` takes its argument positionally; `AuthIdpTheme.defaultTheme()` replaces a snippet that passed a nonexistent field and omitted two required ones.
- Non-compiling `server.dart` snippets across four pages gained their generated imports.

Cross-cutting style: headings sentence-cased, "app" for the Flutter app in prose, brand capitalization, and the remaining semicolon splices split.

`npm run build` passes.


docs/06-concepts/04-authentication/05-providers/07-github/02-customizations.md
```diff
@@ -144,12 +144,12 @@ This callback runs inside the same database transaction as the account creation.
 :::
 
 :::caution
-If you need to assign Serverpod scopes based on provider account data, updating the database alone (via `AuthServices.instance.authUsers.update()`) is **not enough** for the current login session. Token issuance uses the in-memory `authUser.scopes`, which is already set before this callback runs. You would need to update `authUser.scopes` as well. For scope assignment at creation time, use [`onBeforeAuthUserCreated`](#reacting-to-auth-user-creation) in combination with `getExtraGitHubInfoCallback` to fetch and store the data you need before the auth user is created.
+Scopes you assign here with `AuthServices.instance.authUsers.update()` do not apply to the login that is already in progress, because token issuance uses the scopes loaded before this callback runs. They take effect the next time the user signs in. To force them sooner, revoke the user's tokens so they sign in again. The `onBeforeAuthUserCreated` hook cannot help for GitHub, because `getExtraGitHubInfoCallback` runs after the auth user is created.
```
Zfinix: The next section says to use `onBeforeAuthUserCreated` for scopes at creation time, so a flat "cannot help for GitHub" contradicts it. It only fails when the scope depends on GitHub data.

docs/06-concepts/04-authentication/05-providers/07-github/01-setup.md
```diff
@@ -36,14 +36,14 @@ The callback URL is where GitHub redirects the user after they authorize your ap
 
 1. In the **Callback URL** field, enter the redirect URI for your app. GitHub Apps accept up to 10 entries, one per line. Add every platform you target:
 
-   - **iOS and Android**: `com.example.yourapp://auth` (a custom scheme registered in `AndroidManifest.xml` and `Info.plist`).
+   - **iOS and Android**: `com.example.yourapp://auth`, a custom scheme registered in `AndroidManifest.xml`. iOS and macOS need no special configuration.
```
Zfinix: The bullet is labelled iOS and Android but the text also covers macOS, and the only registration step left is Android. Label it iOS, macOS, and Android.

docs/06-concepts/04-authentication/05-providers/06-firebase/02-customizations.md
```diff
@@ -110,7 +122,7 @@ Which properties are populated depends on the Firebase sign-in method the user c
 
 [`onBeforeAuthUserCreated`](https://pub.dev/documentation/serverpod_auth_idp_server/latest/core/AuthUsersConfig/onBeforeAuthUserCreated.html) and [`onAfterAuthUserCreated`](https://pub.dev/documentation/serverpod_auth_idp_server/latest/core/AuthUsersConfig/onAfterAuthUserCreated.html) are global callbacks on `AuthUsersConfig`. They fire for every identity provider, not just Firebase. See [Working with users](../../working-with-users#reacting-to-the-user-created-event) for full details.
 
-The example below uses Firebase phone numbers as the trigger for assigning a `phone-verified` scope at sign-up, and persists the Firebase UID for later admin lookups:
+The core callbacks cannot see Firebase account details, so provider-specific logic, such as a scope derived from the phone number, belongs in `onAfterFirebaseAccountCreated`, which receives the `FirebaseAccount`. The example below assigns a baseline scope to every new user:
```
Zfinix, line 125: This sends the reader to `onAfterFirebaseAccountCreated`, but the example below only shows the core `onBeforeAuthUserCreated` and `onAfterAuthUserCreated`. The callback it points at never appears.

docs/06-concepts/04-authentication/05-providers/04-apple/03-customizing-the-ui.md
```diff
@@ -15,8 +15,12 @@ SignInWidget(
   client: client,
   appleSignInWidget: AppleSignInWidget(
     client: client,
-    // Customize the widget
-    style: AppleButtonStyle.black,
+    // Shape and label survive inside SignInWidget. Brand colors do not.
```
Zfinix: `SignInButtonBase` resolves `shared?.shape ?? this.shape`, so a `buttonStyle` on `SignInWidget` wins over these arguments. Shape and label only survive when no `buttonStyle` is set, which is the opposite of what the UI components page now says. Same comment in the Google, GitHub, and Microsoft UI pages.

Zfinix (changes requested):
Left some comments. The Firebase example and the button style precedence are the blocking ones.

Zfinix (approved):
LGTM

PR #731, opened 2026-08-06
docs: Reshape the Authentication provider pages for 4.0

developerjamiu:
Follow-up to #728 and #729. This PR normalizes the shape of the Authentication provider pages so every provider group follows the same structure. Content changes are limited to what the restructure requires plus fixes found in the pre-PR review.

## Structure

- **One customizations page per OAuth group.** For anonymous, Google, Apple, Facebook, GitHub, and Microsoft, the old configuration/customizations page and the separate "Customizing the UI" page are merged into a single `customizations` page (server configuration, then app configuration, then UI). Email and Firebase keep their separate UI pages since their scope differs.
- **Microsoft brought up to group parity.** Its `configuration` page is renamed to `customizations`, and it gains a troubleshooting page matching its siblings (16 entries, each verified against the framework source).
- **Troubleshooting pages renumbered to 03** across the merged groups, now that the UI pages are gone.
- **Passkey UI stub removed.** Its content already lives on the setup page.
- **Custom providers overview rewritten** from a stub into a real entry page (what a custom provider consists of, OAuth2 path vs everything else).
- **auth.html has a single home** on the core setup page; other mentions now point there.

## Content fixes from the review pass

- Google web setup used the `/app` serving path throughout; default projects (no website option) serve the built app at `/`. Corrected in four places, with the website-option case called out.
- The `google-services.json` fallback on Android only works with the Firebase Gradle plugin; plain projects must pass client IDs in code or with `--dart-define`.
- The anonymous rate-limit snippet now shows the import that `RateLimit` actually requires.
- Facebook gained the credential-loading section its setup page promised, plus a `FacebookIdpConfig` parameter reference matching the Google/GitHub tables.
- The Apple email/name persistence story is now consistent between setup and troubleshooting (Serverpod stores the values automatically on first sign-in and does not backfill them later).
- Smaller flow and wording fixes across the group: consistent Related blocks on all five troubleshooting pages, accurate headings for the missing-password startup crash, and cross-links between the split scopes/server-API sections.

## Redirects

The removed URLs (`customizing-the-ui` pages, `microsoft/configuration`, `anonymous/configuration`) only exist in the `/next` version, so `@docusaurus/plugin-client-redirects` rejects redirect entries for them while 3.4.0 is the served default. The full redirect list is recorded and will land at the 4.0 GA cutover.

PR #732, opened 2026-08-07
docs: Correct the legacy Authentication pages for 4.0

developerjamiu:
Final PR of the Authentication section pass, following #728, #729, and #731. This one covers the legacy section, the two upgrading guides, and two small consistency follow-ups from #731.

## Legacy pages (correctness only, no restructuring)

Every claim on the nine legacy pages was verified against the framework source. The legacy `serverpod_auth` module still ships at 4.0.0-beta.1 (relocated to `modules/legacy/` in the monorepo), so the pages are corrected to match that release:

- **Removed APIs are no longer documented as callable.** The `signOut` method and status endpoint were removed in 3.0; the info boxes now say so and point at `signOutDevice`/`signOutAllDevices`. The `legacyUserSignOutBehavior` config row is gone. Client snippets assign the key manager to the `authKeyProvider` field instead of the removed constructor parameter.
- **Non-compiling snippets fixed:** `AuthenticationInfo` calls (String identifier, required `authId`), `UserAuthentication.signInUser` (missing Session argument), the `passwordHashValidator`/`passwordHashGenerator` examples (current signature and `PasswordValidationResult`), the Apple button import (pointed at the email package), stale `await`s on the now-synchronous `session.authenticated` and `isUserSignedIn` getters, and import standardization on the `serverpod_auth_server.dart` entrypoint.
- **Wrong defaults corrected:** `passwordResetExpirationTime` is 15 minutes (was documented as 24h), `allowUnsecureRandom` defaults to the safe behavior (the page described the opposite), `validationCodeLength` documents the real warn-below-8/throw-below-4 behavior, and the firebase key path and image-generator defaults are filled in.
- **Version pins:** `^1.x.x` pins replaced with 4.0.0-beta.1 across the section.
- **Positioning and migration linking:** the setup intro now presents the module as the legacy option for existing apps, points new projects to the current auth setup, and links the migration guide. The other eight pages carry a uniform migration note under the H1.

## Upgrading guides

- **Upgrade to 4.0** gains an "If you use the legacy auth module" subsection in the dependencies step: bump the `serverpod_auth` packages in lockstep, and move off the deprecated `authenticationKeyManager` parameter to `authKeyProvider` (deprecated at beta.1, removed after).
- Both upgrading pages now pin **4.0.0-beta.1** consistently (the migration guide previously told a freshly upgraded reader to downgrade back to beta.0), and the migration guide's SDK prerequisites match the beta.1 minimums (Dart 3.10.3, Flutter 3.38.4).

## Follow-ups from #731

- **Firebase's configuration page is renamed** from "Customizations" to "Configuration" (file, sidebar label, and H1), matching email, so "Customizations" consistently means the merged config-plus-UI scope across providers. The old URL's redirect is queued for the GA cutover with the rest.
- **Platform casing:** "web" is lowercased in platform lists and prose across the section, while console UI labels, Google's "Web application" OAuth client type, and sentence-initial uses stay capitalized.


Zfinix (commented):
`legacyUserSignOutBehavior` is still on `AuthConfig` in 4.0.0-beta.1, deprecated not removed. Should stay in the table with a deprecation note, same as `authenticationKeyManager` in the upgrade guide.

Zfinix (commented):
Nit: `passwordResetExpirationTime` doesn't say what unit, only the Default column does.

Zfinix (commented):
The Firebase customizations rename has no redirect, the old URL 404s until GA.

Zfinix (approved):

Zfinix (approved):
LGTM 🚀

developerjamiu:
> `legacyUserSignOutBehavior` is still on `AuthConfig` in 4.0.0-beta.1, deprecated not removed.

`legacyUserSignOutBehavior` was removed in 3.0.0 together with the `signOut` endpoint (#4084): `git grep legacyUserSignOutBehavior 4.0.0-beta.1` returns zero hits repo-wide, while 2.9.3 still has it. Unlike `authenticationKeyManager`, no deprecated shim exists at beta.1, so the row stays out.

developerjamiu:
> Nit: `passwordResetExpirationTime` doesn't say what unit, only the Default column does.

Applied.

developerjamiu:
> The Firebase customizations rename has no redirect, the old URL 404s until GA.

The redirect plugin validates targets against the served 3.4.0 version, where the new URL doesn't exist, so the build rejects the entry. It's queued for the GA cutover together with the other reshape redirects.

PR #739, opened 2026-08-14
docs: Document using models on the client only

### LINKED ISSUE #3670: Mention creating models that are used client-side only
### Summary of the issue

It's currently not possible to create models that are client-side only. However, there is no harm in creating models that are only used by the client. We should mention this in the documentation.

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [ ] I want to work on this

abdulawalarif:
## Summary

Working with models covered `serverOnly` and field `scope`, but not that you can define models you only use in Flutter. There is no `clientOnly` flag. Place the YAML in the server package as usual; generating the class on the server as well is harmless.

Fixes https://github.com/serverpod/serverpod/issues/3670

## Changes

- Add a "Using models on the client only" section after the visibility examples.
- Show placing the `.spy.yaml` file in the server package, running code generation, and importing the class from the client package.
- Note that omitting `table` avoids creating a database table, and point to `immutable: true` and shared packages.

## Reference

- serverpod/serverpod#3670

### DIFF
--- a/docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
+++ b/docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
@@ -104,23 +104,22 @@ Models can be saved to and read from the database. See the [Database](./database
 
 You can define models that you only use in the Flutter app, for example to hold local UI state or form data. Leave `serverOnly` unset so the class is generated for the client. There is no separate `clientOnly` flag.
 
-Place the `.spy.yaml` file anywhere in the server's `lib` directory, run code generation, and import the class from the client package.
+Place the `.spy.yaml` file anywhere in the server's `lib` directory. With `serverpod start` running, saving the file regenerates the code; otherwise run `serverpod generate`. Then import the class from the client package.
 
 ```yaml
 class: CartItem
-immutable: true
 fields:
   productId: String
   quantity: int
 ```
 
 ```dart
-import 'package:my_project_client/my_project_client.dart';
+import 'package:your_client/your_client.dart';
 
 var item = CartItem(productId: 'sku-1', quantity: 2);
 ```
 
-The class is also generated on the server. That is harmless: omit the `table` key so no database table is created, and you do not need to use the class in any endpoint. Set `immutable: true` if you want value equality for Flutter state management. See [Immutable classes](#immutable-classes).
+The class is also generated on the server, which is harmless: without a `table` key no database table is created, and nothing requires you to use the class in an endpoint. Set `immutable: true` if you want value equality for Flutter state management. See [Immutable classes](#immutable-classes).
 
 If you want the YAML and generated classes in a package that both the server and the app depend on, use a [shared package](./models/shared-packages).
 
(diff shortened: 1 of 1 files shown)


docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
```diff
@@ -100,6 +100,30 @@ fields:
...
 
+### Using models on the client only
+
+You can define models that you only use in the Flutter app, for example to hold local UI state or form data. Leave `serverOnly` unset so the class is generated for the client. There is no separate `clientOnly` flag.
+
+Place the `.spy.yaml` file anywhere in the server's `lib` directory, run code generation, and import the class from the client package.
+
+```yaml
+class: CartItem
+immutable: true
```
developerjamiu: Suggestion: drop `immutable: true` from this example. The page only introduces immutability further down, and the sentence below the example already points readers there when they need value equality.

```suggestion
```

docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
```diff
@@ -100,6 +100,30 @@ fields:
 Models can be saved to and read from the database. See the [Database](./database/tables) section.
 :::
 
+### Using models on the client only
+
+You can define models that you only use in the Flutter app, for example to hold local UI state or form data. Leave `serverOnly` unset so the class is generated for the client. There is no separate `clientOnly` flag.
+
+Place the `.spy.yaml` file anywhere in the server's `lib` directory, run code generation, and import the class from the client package.
```
developerjamiu: Suggestion: mirror how the intro of this page describes code generation, so both spots teach the same loop.

```suggestion
Place the `.spy.yaml` file anywhere in the server's `lib` directory. With `serverpod start` running, saving the file regenerates the code; otherwise run `serverpod generate`. Then import the class from the client package.
```

docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
```diff
@@ -100,6 +100,30 @@ fields:
...
+```yaml
+class: CartItem
+immutable: true
+fields:
+  productId: String
+  quantity: int
+```
+
+```dart
+import 'package:my_project_client/my_project_client.dart';
```
developerjamiu: Suggestion: the other import examples in the docs use `your_client`, so keeping that name here helps readers connect them.

```suggestion
import 'package:your_client/your_client.dart';
```

docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
```diff
@@ -100,6 +100,30 @@ fields:
...
+
+var item = CartItem(productId: 'sku-1', quantity: 2);
+```
+
+The class is also generated on the server. That is harmless: omit the `table` key so no database table is created, and you do not need to use the class in any endpoint. Set `immutable: true` if you want value equality for Flutter state management. See [Immutable classes](#immutable-classes).
```
developerjamiu: Suggestion: this reads as if omitting `table` is what makes the server copy harmless; the two facts sit side by side rather than cause and effect.

```suggestion
The class is also generated on the server, which is harmless: without a `table` key no database table is created, and nothing requires you to use the class in an endpoint. Set `immutable: true` if you want value equality for Flutter state management. See [Immutable classes](#immutable-classes).
```

abdulawalarif:
Thanks for the review. 
Applied all four suggestions.

developerjamiu (approved):

developerjamiu:
Thanks for the contribution

PR #740, opened 2026-08-14
docs: Render the Cloud Dart SDK versions from a data file

Isakdl:
# Changes

Moves the supported Dart SDK version list into a data file, `cloud_docs/reference/dart-sdk-versions.json` (supported versions ordered lowest to highest, plus the default), and rewrites the page as MDX that renders from it. The page stays at its existing URL, `/cloud/reference/dart-sdk-versions`; the URL mismatch behind CLD-558 (the server and CLI print a `guides/deployment` URL that 404s) is fixed from the cloud side instead, in serverpod/serverpod_cloud#2380.

A reconciliation loop in `serverpod/serverpod_cloud` (the daily `Dart version bump` workflow, serverpod/serverpod_cloud#2377) will keep the data file in sync with the version policy the prod platform endpoint serves, opening auto-merged PRs here on drift — only the JSON changes on those PRs, the prose stays put. This PR also allowlists that sync branch (`auto/dart-versions-sync`) in the auto-approve workflow so those PRs can merge unattended.

The rewrite also fixes two content errors: the supported list was missing 3.11, and the page claimed the lowest overlapping version is used for builds when the platform picks the highest. It now also documents how to select a version on deploy (`--dart-version` flag, `scloud.yaml`, `.tool-versions`, pubspec constraint, in that precedence order), which was the other half of CLD-558.

Linear: https://linear.app/serverpod/issue/CLD-1197/docs-render-the-dart-version-list-from-data-updated-by-reconciliation

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cloud_docs/reference/dart-sdk-versions.mdx
```diff
@@ -0,0 +1,75 @@
...
+    <li key={version}>
+      Dart {version}
+      {version === versionData.defaultVersion ? ' (default)' : ''}
+    </li>
+  ))}
+</ul>
+
+## How the version is selected
+
+`scloud deploy` reads the Dart SDK version from the first of these sources that sets one:
```
developerjamiu: Suggestion: lead with a noun phrase so the sentence doesn't open on a symbol, matching how the docs introduce commands elsewhere.

```suggestion
The `scloud deploy` command reads the Dart SDK version from the first of these sources that sets one:
```

cloud_docs/reference/dart-sdk-versions.json
```diff
@@ -0,0 +1,9 @@
+{
+  "supportedVersions": [
+    "3.8",
+    "3.9",
+    "3.10",
+    "3.11"
```
developerjamiu, line 6: Question: in the `serverpod_cloud` repo, `dart_image_tags.dart` on main also lists `3.12` as supported, but this file stops at 3.11. Is that intentional because production doesn't offer 3.12 yet? If so, I assume the daily sync workflow will add it here once it ships.

Isakdl: Correct! the 3.12 is queued for deployment but not yet live

Isakdl: the automated workflow should come with an update! I will give it a manual trigger as soon as all the parts are merged.

developerjamiu (approved):
LGTM

PR #743, opened 2026-08-18
docs: Document the generated foreign key field for relations

marcelomendoncasoares:
Documents the model changes from serverpod/serverpod#5596, which generates the foreign key field named by `relation(field=...)` when the model does not declare it, and lifts the mutual exclusion between `field` and `optional`.

- **Custom foreign key field** rewritten: the named field is generated with the id type of the related model, `optional` makes it nullable, and declaring it yourself is only needed for a column override or a scope.
- **Bidirectional relations** example no longer declares `addressId`.
- Model reference note for `field` updated.


developerjamiu:
Suggestion: `04-self-relations.md:19-20` still declares a foreign key that this PR documents as generated:

```yaml
  nextId: int?
  next: Post?, relation(name=next_previous_post, field=nextId, onDelete=SetNull)
```

`nextId` has no `column` override and no scope, so it is the case the rewrite says no longer needs declaring.

Not a plain deletion though: `nextId` is nullable while `next` is not `optional`, so dropping the line generates a non-nullable field and turns the column NOT NULL. It would need `optional` added to the relation. Fold it in here or take a follow-up, whichever you prefer.

docs/06-concepts/03-data-and-the-database/02-database/06-relations/01-one-to-one.md
```diff
@@ -90,24 +90,36 @@ With the introduction of the `optional` keyword in the relation, the automatical
...
+
+Declare the field yourself when you need control over it, for example to give it a [column name override](../tables#column-name-override) or a [scope](../../models#limiting-visibility-of-a-generated-class). A declared field must be nullable if the relation is `optional`:
+
+```yaml
+# user.yaml
+class: User
+table: user
+fields:
+  customIdField: int?, column=fk_user_address_id, unique
+  address: Address?, relation(optional, field=customIdField)
```
developerjamiu, line 122: Suggestion: this example expresses uniqueness with the field-level `unique` keyword, while the three examples above it use an `indexes:` block. `01-tables.md:239-244`, which this sentence links to for the column override, also uses the block form, so a reader who follows that link lands on a different shape for the same job.

```suggestion
# user.yaml
class: User
table: user
fields:
  customIdField: int?, column=fk_user_address_id
  address: Address?, relation(optional, field=customIdField)
indexes:
  user_address_unique_idx:
    fields: customIdField
    unique: true
```

docs/06-concepts/03-data-and-the-database/02-database/06-relations/01-one-to-one.md
```diff
@@ -90,24 +90,36 @@ With the introduction of the `optional` keyword in the relation, the automatical
...
     fields: customIdField
     unique: true
 ```
 
-In this example, we define a custom foreign key field with the `field` parameter. The argument defines what field that is used as the foreign key field. In this case, `customIdField` is used instead of the default auto-generated name.
+In this example, `customIdField` holds the foreign key instead of the default auto-generated name `addressId`. The field does not have to be declared in `fields`. Serverpod generates it with the id type of the related model, just like it does for a relation without the `field` parameter, and the `optional` keyword makes it nullable:
 
-If you want the custom foreign key to be nullable, simply define its type as `int?`. Note that the `field` keyword cannot be used in conjunction with the `optional` keyword. Instead, directly mark the field as nullable.
+```yaml
+  address: Address?, relation(optional, field=customIdField)
```
developerjamiu, line 111: Nit: this snippet is a single indented line with no surrounding keys, unlike every other example on the page. Adding the `fields:` key gives it enough context to read on its own.

```suggestion
fields:
  address: Address?, relation(optional, field=customIdField)
```

marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod_docs/pull/743/commits/18682ab08497f81b798d4b971a12ad94c65899d7.

docs/06-concepts/03-data-and-the-database/02-database/06-relations/01-one-to-one.md
marcelomendoncasoares: Made more consistent at https://github.com/serverpod/serverpod_docs/pull/743/commits/18682ab08497f81b798d4b971a12ad94c65899d7.

developerjamiu (approved):

PR #744, opened 2026-08-18
docs: Replace the global flag on postMessage with MessageScope

marcelomendoncasoares:
The `global` parameter on `postMessage` was replaced by `scope` in serverpod/serverpod#5553, and the default is now `MessageScope.auto`: cluster-wide when Redis is enabled, local otherwise.

Rewrites the section as **Message scope** with a table of the three values, and updates the inbound link from the streaming page.

### DIFF
--- a/docs/06-concepts/02-endpoints-and-apis/05-server-events.md
+++ b/docs/06-concepts/02-endpoints-and-apis/05-server-events.md
@@ -17,22 +17,24 @@ await session.messages.postMessage('user_updates', message);
 
 In the example above, the message is published on the `user_updates` channel. Any subscriber to this channel in the server will receive the message.
 
-### Global messages
+### Message scope
 
-Serverpod uses Redis to pass messages between servers. To send a message to another server, [enable Redis](../server-fundamentals/configuration) and then set the `global` parameter to `true` when posting a message.
+Serverpod uses Redis to pass messages between servers. A message is sent to every server connected to the same Redis instance when [Redis is enabled](../server-fundamentals/configuration), and stays on the current server when it is not. Set the `scope` parameter to choose the delivery explicitly.
 
 ```dart
 var message = UserUpdate(); // Model that represents changes to user data.
-await session.messages.postMessage('user_updates', message, global: true);
+await session.messages.postMessage(
+  'user_updates',
+  message,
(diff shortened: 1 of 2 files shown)


developerjamiu:
Suggestion: `05-server-events.md:79` still uses the old vocabulary:

> Listeners receive both local and global messages.

It falls outside the diff, so the page now names `MessageScope.global` as a scope in one section and "global messages" as a kind of message in another.

Suggested: "Listeners receive every message that reaches this server, whatever its scope."

docs/06-concepts/02-endpoints-and-apis/05-server-events.md
```diff
@@ -17,22 +17,24 @@ await session.messages.postMessage('user_updates', message);
...
+Serverpod uses Redis to pass messages between servers. A message is sent to every server connected to the same Redis instance when [Redis is enabled](../server-fundamentals/configuration), and stays on the current server when it is not. Set the `scope` parameter to choose the delivery explicitly.
 
 ```dart
 var message = UserUpdate(); // Model that represents changes to user data.
-await session.messages.postMessage('user_updates', message, global: true);
+await session.messages.postMessage(
+  'user_updates',
+  message,
+  scope: MessageScope.global,
+);
```
developerjamiu, line 38: Suggestion: the sample uses `MessageScope.global` before the table below introduces the three values, and it picks the one value that throws right after the prose has said the default already goes cluster-wide when Redis is on, so nothing motivates the choice.

Moving the table above the sample fixes the forward reference. `MessageScope.local` is also the more instructive value here, since it is the one case the default will not give you. If `global` stays, a clause on why you would force it (fail loudly rather than degrade to local) would help.

developerjamiu:
> Suggestion: `05-server-events.md:79` still uses the old vocabulary:
> 
> > Listeners receive both local and global messages.
> 
> It falls outside the diff, so the page now names `MessageScope.global` as a scope in one section and "global messages" as a kind of message in another.
> 
> Suggested: "Listeners receive every message that reaches this server, whatever its scope."

Saw this @marcelomendoncasoares ?

developerjamiu (approved):
LGTM

developerjamiu (approved):

PR #746, opened 2026-08-18
docs: Document table models in shared packages

marcelomendoncasoares:
Shared packages can define table models since serverpod/serverpod#5439, provided the model also sets `database: all`. The page still said they cannot.

Adds a short **Table models** section and corrects the restriction row.


developerjamiu (approved):
LGTM

PR #747, opened 2026-08-18
docs: Correct the migrate shortcut in the serverpod start terminal

marcelomendoncasoares:
Since serverpod/serverpod#5481, **M** creates the migration and applies it in one step, and the same holds for **P**. The **A** shortcut remains as a retry when applying fails.

Updates the six places that still tell the reader to press **M** then **A**.

### DIFF
--- a/docs/06-concepts/01-server-fundamentals/02-running-your-server.md
+++ b/docs/06-concepts/01-server-fundamentals/02-running-your-server.md
@@ -32,7 +32,7 @@ On the first boot of a session, `serverpod start` applies any pending migrations
 
 - **M** creates a migration from your current model changes and applies it.
 - **P** creates a repair migration to reconcile a database that has drifted from your migrations, and applies it. This shortcut is not shown in the bottom bar; press **H** to see it listed.
-- **A** applies pending migrations, to retry after a failed apply.
+- **A** applies pending migrations. Use it to retry after a failed apply.
 
 Hold **Shift** with **M** or **P** to force the migration: it is created even when there are no changes to record, and even when Serverpod warns that information may be destroyed, so force deliberately. For how migrations work, see [Migrations](../data-and-the-database/database/migrations).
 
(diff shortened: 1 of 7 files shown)


developerjamiu:
Suggestion: `05-build-your-first-app/03-working-with-the-database.md:48-49` still gives the two-step instruction, under a heading called "Create and apply the migration":

```
- Press **M** to create the migration from your model change.
- Press **A** to apply it, which creates the `recipes` table in your database.
```

It falls outside the diff, and it is where a new user meets the TUI for the first time. Suggested replacement for both bullets:

```
- Press **M** to create the migration and apply it, which creates the `recipes` table in your database.
```

docs/06-concepts/01-server-fundamentals/02-running-your-server.md
```diff
@@ -30,9 +30,9 @@ To start without watching, pass `--no-watch`. Nothing reloads on save; there, **
 
 On the first boot of a session, `serverpod start` applies any pending migrations. While it runs, you handle further schema changes from the terminal without leaving the session:
 
-- **M** creates a migration from your current model changes.
-- **A** applies pending migrations to the database.
-- **P** creates a repair migration to reconcile a database that has drifted from your migrations. This shortcut is not shown in the bottom bar; press **H** to see it listed.
+- **M** creates a migration from your current model changes and applies it.
+- **P** creates a repair migration to reconcile a database that has drifted from your migrations, and applies it. This shortcut is not shown in the bottom bar; press **H** to see it listed.
+- **A** applies pending migrations, to retry after a failed apply.
```
developerjamiu: Suggestion: this bundles two ideas behind a comma. `02-setup.md:133` phrases the same thing well, "The migration is created and applied in one step. If applying fails, press **A** to retry it."

```suggestion
- **A** applies pending migrations. Use it to retry after a failed apply.
```

The same reads awkwardly at `13-migrations.md:115`, where "retry applying ... if applying them failed" repeats the verb. "Press **A** to retry when applying a migration fails." would work there.

marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod_docs/pull/747/commits/b83940a908e11ea3a49904d645569a246d1e03f7, together with the other place that is out of diff.

developerjamiu (approved):

PR #749, opened 2026-08-18
docs: State that the legacy streaming API is removed in 4.0

marcelomendoncasoares:
serverpod/serverpod#5544 removed the legacy streaming session and the deprecated streaming APIs. The streaming page still called them deprecated, and the session type table still listed `StreamingSession`, which no longer exists.

Rewords the section on the streaming page and drops the row from the session type table. The archive page keeps documenting the old API for readers on 3.x.

### DIFF
--- a/docs/11-upgrading/01-upgrade-to-four.md
+++ b/docs/11-upgrading/01-upgrade-to-four.md
@@ -84,6 +84,12 @@ Then refresh the generated server and client code:
 $ serverpod generate
 ```
 
+## If you use the legacy streaming endpoints API
+
+Serverpod's legacy streaming endpoints API was deprecated in 3.0 and is removed in 4.0. Endpoints that use the `StreamingSession` type no longer compile, and all the related server and client methods (e.g. `streamOpened`, `streamClosed`, `handleStreamMessage`, `sendStreamMessage`, `getUserObject`, `setUserObject`, `openStreamingConnection`) are gone.
+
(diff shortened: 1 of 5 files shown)


developerjamiu:
Suggestion: `02-logging.md:41-50` documents a log table that serverpod/serverpod#5544 stopped filling. `logMessage()` had a single caller, in `endpoint_websocket_request_handler.dart` right after `handleStreamMessage`, and that file is gone.

The count and the cascade sentence go with the row:

```
In the database they land in three tables:

| Table | Holds |
| --- | --- |
| `serverpod_session_log` | One row per completed session. |
| `serverpod_log` | Your `session.log` messages. |
| `serverpod_query_log` | Database queries. |

The last two reference the session row, so deleting a session row removes its queries with it.
```

The model and the Insights read stay, so old rows are still queryable. Nothing on 4.0 writes new ones.

docs/11-upgrading/01-upgrade-to-four.md
```diff
@@ -84,6 +84,12 @@ Then refresh the generated server and client code:
 $ serverpod generate
 ```
 
+## If you use the legacy streaming endpoints API
```
developerjamiu: Should be H3 same level as `### If you use the legacy auth module`.

marcelomendoncasoares: Done at https://github.com/serverpod/serverpod_docs/pull/749/commits/07f91536d7d71e54ed953c6a6b0699232bd90b2b (this and the removal of the stream message logs).

I decided to completely remove the table mention instead of leaving a dangling reference to the legacy since it is already deprecated for a long time and it is relevant for logs only.

developerjamiu (approved):

PR #752, opened 2026-08-18
docs: Document nulls_distinct on unique indexes

marcelomendoncasoares:
Documents the `nulls_distinct` index key added in serverpod/serverpod#5547, with the "one active row per key" case it exists for and the SQLite limitation. Adds the keyword to the model reference.


docs/06-concepts/03-data-and-the-database/02-database/09-indexing.md
```diff
@@ -99,6 +99,28 @@ indexes:
...
+indexes:
+  product_active_sku_idx:
+    fields: tenantId, sku, deletedAt
+    unique: true
+    nulls_distinct: false
+```
+
+In this example, a tenant can have at most one product with a given `sku` and no `deletedAt`. Deleted rows carry a timestamp, so they no longer collide with each other.
+
+The key is only allowed on unique indexes, and leaving it out keeps the database default. SQLite cannot express `nulls_distinct: false`, so creating a migration for SQLite with models that use it will fail.
```
developerjamiu: Suggestion: "models that use it" reads as "models that use `nulls_distinct`", but only `false` throws. `sqlite.dart:220` guards on `nullsDistinct == false`, so `nulls_distinct: true` is accepted and emits a plain unique index, which is already SQLite's behaviour.

```suggestion
The key is only allowed on unique indexes, and leaving it out keeps the database default. SQLite cannot express `nulls_distinct: false`, so creating a SQLite migration for a model that sets it fails. Setting it to `true` is accepted, since that matches what SQLite already does.
```

developerjamiu (dismissed):

docs/06-concepts/03-data-and-the-database/02-database/09-indexing.md
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod_docs/pull/752/commits/ceaed9c0493c1a58ab7436a11617cd4836b9fce0.

developerjamiu (approved):

PR #753, opened 2026-08-18
docs: Document defaultPersist=serial on int fields

marcelomendoncasoares:
serverpod/serverpod#5556 allows `serial` on any `int` column, not only `id`. The default values section did not mention `serial` at all.

### DIFF
--- a/docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
+++ b/docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
@@ -233,13 +233,13 @@ For `UuidValue`, `random` generates a UUID v4 (`Uuid().v4obj()` in Dart, `gen_ra
 
 For enums, the persisted value follows the enum's serialization mode. A `byName` enum default is stored as the value's name, and a `byIndex` enum default is stored as its index, e.g. `0` for the first value.
 
-On an `int` field, `defaultPersist=serial` hands the value over to a database sequence (same used by an `int` `id` column). Because only the database can produce it, `serial` is rejected with `default` and `defaultModel`. On SQLite, it is only supported on the `id` column due to dialect limitations.
+On an `int` field, `defaultPersist=serial` hands the value over to a database sequence, the same mechanism behind an `int` `id` column, though each serial column gets its own sequence. Because only the database can produce it, `serial` is rejected with `default` and `defaultModel`. On SQLite, it is only supported on the `id` column due to dialect limitations.
 
 ```yaml
 class: Invoice
 table: invoice
 fields:
-  invoiceNumber: int, defaultPersist=serial
+  invoiceNumber: int?, defaultPersist=serial
 ```
 
 :::info
@@ -259,7 +259,7 @@ fields:
   boolDefault: bool, defaultModel=false, defaultPersist=true
 
   ### Sets the database-side default value for an integer field.
-  intDefault: int, defaultPersist=20
+  intDefault: int?, defaultPersist=20
 
   ### Defaults to 10.5 in code, with a separate database default.
   doubleDefault: double, default=10.5, defaultPersist=20.5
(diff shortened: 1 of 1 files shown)


docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
```diff
@@ -233,6 +233,15 @@ For `UuidValue`, `random` generates a UUID v4 (`Uuid().v4obj()` in Dart, `gen_ra
 
 For enums, the persisted value follows the enum's serialization mode. A `byName` enum default is stored as the value's name, and a `byIndex` enum default is stored as its index, e.g. `0` for the first value.
 
+On an `int` field, `defaultPersist=serial` hands the value over to a database sequence (same used by an `int` `id` column). Because only the database can produce it, `serial` is rejected with `default` and `defaultModel`. On SQLite, it is only supported on the `id` column due to dialect limitations.
+
+```yaml
+class: Invoice
+table: invoice
+fields:
+  invoiceNumber: int, defaultPersist=serial
```
developerjamiu: Suggestion: this example does not validate. `restrictions.dart:2786` errors when a field has only a `defaultPersist` and is non-nullable: "When setting only the \"defaultPersist\" key, its type should be nullable". The framework's passing fixture for this feature uses `syncVersion: int?, defaultPersist=serial`.

```suggestion
class: Invoice
table: invoice
fields:
  invoiceNumber: int?, defaultPersist=serial
```

docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
```diff
@@ -233,6 +233,15 @@ For `UuidValue`, `random` generates a UUID v4 (`Uuid().v4obj()` in Dart, `gen_ra
 
 For enums, the persisted value follows the enum's serialization mode. A `byName` enum default is stored as the value's name, and a `byIndex` enum default is stored as its index, e.g. `0` for the first value.
 
+On an `int` field, `defaultPersist=serial` hands the value over to a database sequence (same used by an `int` `id` column). Because only the database can produce it, `serial` is rejected with `default` and `defaultModel`. On SQLite, it is only supported on the `id` column due to dialect limitations.
```
developerjamiu: Suggestion: "same used by an `int` `id` column" is wrong on either reading. As "the same sequence" it is false, since each serial column gets its own. As "the same mechanism" it still misleads, because `postgres.dart:258` emits `bigserial` for the primary key and plain `serial` otherwise, so this column holds about 2.1 billion values where `id` holds far more.

```suggestion
On an `int` field, `defaultPersist=serial` hands the value to a database sequence, the same mechanism behind an `int` `id` column. The column is created as `serial` rather than the `bigserial` used for `id`, so it holds about 2.1 billion values. Because only the database can produce the value, `serial` is rejected with `default` and `defaultModel`. On SQLite, it is only supported on the `id` column due to dialect limitations.
```

Drop the range sentence if that is more detail than the section wants.

developerjamiu:
Comment: the same rule catches an existing example two blocks below, at line 262:

```yaml
  ### Sets the database-side default value for an integer field.
  intDefault: int, defaultPersist=20
```

Only a `defaultPersist` and a non-nullable type, so it fails the same check. It matches the framework's error fixture, `intType: int, defaultPersist=10`, almost character for character.

Pre-existing rather than introduced here, but it is the same page and the same keyword this PR documents.

docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod_docs/pull/753/commits/5f54be7ae506575528bf7703a3f6180fe08c5c67, together with the other error that already existed.

docs/06-concepts/03-data-and-the-database/01-models/01-working-with-models.md
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod_docs/pull/753/commits/5f54be7ae506575528bf7703a3f6180fe08c5c67.

developerjamiu (approved):

PR #755, opened 2026-08-18
docs: Document cache-busted asset URLs in HTML templates

marcelomendoncasoares:
serverpod/serverpod#5444 added the `{{{@/path}}}` template tag and the `cacheBustingConfig` parameter on `WidgetRoute`. Cache busting was only documented for `StaticRoute` and the manual `assetPath()` call.


marcelomendoncasoares:
@Crazelu Can you check this PR before we request a review from @serverpod/docs? Just to be sure it is correct on our intention for how people should use the feature.

Crazelu:
> @Crazelu Can you check this PR before we request a review from @serverpod/docs? Just to be sure it is correct on our intention for how people should use the feature.

Checked. It is correct 👍

developerjamiu (approved):

PR #758, opened 2026-08-18
docs: Use the Serverpod PostgreSQL image for PostGIS support

marcelomendoncasoares:
serverpod/serverpod#5563 publishes `ghcr.io/serverpod/postgres:16` with both PostGIS and pgvector installed, and new projects reference it in their `docker-compose.yaml`.

Replaces the `postgis/postgis` image and the hand-built Dockerfile for projects that need both extensions.

### DIFF
--- a/docs/11-upgrading/04-upgrade-to-postgis.md
+++ b/docs/11-upgrading/04-upgrade-to-postgis.md
@@ -16,12 +16,12 @@ If trying to use geography fields without upgrading, you will encounter an error
 
 ## For Docker-based environments
 
-1. Update your `docker-compose.yml` to use a PostgreSQL image with PostGIS (e.g., `postgis/postgis:16-3.5`):
+1. Update your `docker-compose.yml` to use `ghcr.io/serverpod/postgres:16`, the Serverpod PostgreSQL image. It ships with [PostGIS and pgvector](../concepts/data-and-the-database/database/vector-and-geography-fields) already installed:
 
 ```yaml
 services:
   postgres:
-    image: postgis/postgis:16-3.5  # <-- Change from postgres image here
+    image: ghcr.io/serverpod/postgres:16  # <-- Change from postgres image here
     ports:
       - '8090:5432'
     environment:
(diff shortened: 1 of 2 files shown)


docs/11-upgrading/04-upgrade-to-postgis.md
```diff
@@ -45,32 +45,7 @@ services:
...
-    ports:
-      - '8090:5432'
-    environment:
-      POSTGRES_USER: postgres
-      POSTGRES_DB: <projectname>
-      POSTGRES_PASSWORD: <DB_PASSWORD>
-    volumes:
-      - <projectname>_data:/var/lib/postgresql/data
-```
+Projects created with Serverpod 4.0 already reference this image, so there is nothing to change for them.
```
developerjamiu: Suggestion: this is not accurate on two counts.

The standard new-project template has no compose file. Only `modulename_server` and `projectname_server_upgrade` have one; `projectname_server` has none, since 4.0 defaults to the embedded Postgres.

And the image landed in serverpod/serverpod#5563 on 2026-08-10, which is in 4.0.0-beta.3 but not beta.1 or beta.2. Anyone who scaffolded on an earlier 4.0 beta has the old image and does need the change this sentence tells them to skip.

```suggestion
If your `docker-compose.yaml` came from Serverpod 4.0.0-beta.3 or later, it already references this image and needs no change. New projects default to the embedded Postgres and have no compose file at all.
```

developerjamiu:
Comment: `02-archive/05-upgrade-to-pgvector.md:3` still opens with a present-tense claim about the old image:

> New Serverpod projects automatically include pgvector support through the `pgvector/pgvector` PostgreSQL Docker image.

The `pgvector/pgvector:pg16` references further down that page are fine, since they describe that upgrade as it was, but this line is about today.

It is an archive page, so leaving it is reasonable. Raising it because it is the same image change.

docs/11-upgrading/04-upgrade-to-postgis.md
marcelomendoncasoares: Fixed the beta reference, but new templates actually have the `docker-compose.yaml` despite using the embedded Postgres. We generate it because users might need to switch, or to test with an actual Redis, and then the composer is needed.

developerjamiu (approved):

PR #759, opened 2026-08-18
docs: Use the generated Serverpod class in initialization examples

FXschwartz:
Updates the documentation for the simplified `Serverpod` initialization introduced in serverpod/serverpod#5611.

`serverpod generate` now produces a `lib/src/generated/serverpod.dart` with a `Serverpod` class pre-wired with the project's `Protocol` and `Endpoints`, so `server.dart` creates the server with just:
```dart
import 'src/generated/serverpod.dart';
final pod = Serverpod(args);
```


docs/06-concepts/01-server-fundamentals/01-your-serverpod-project.md
```diff
@@ -71,20 +71,21 @@ The scaffolded project follows the same pattern beyond the greeting example: `li
...
 
-The server starts in `bin/main.dart`, which only calls the `run` function in `lib/server.dart`. That function creates the server, wires it to your generated code, registers everything that is not an endpoint, and starts it:
+The server starts in `bin/main.dart`, which only calls the `run` function in `lib/server.dart`. That function creates the server from the generated `Serverpod` class, registers everything that is not an endpoint, and starts it:
 
 ```dart
 void run(List<String> args) async {
-  // Connect the server with your generated code.
-  final pod = Serverpod(args, Protocol(), Endpoints());
+  // The generated Serverpod class comes pre-connected with your
+  // generated code.
```
marcelomendoncasoares: Suggestion: The "pre-connected" only make sense thinking on the `Protocol` and `Endpoint` classes, that new users will no longer know exist. We can simplify this comment.

```suggestion
  // The generated Serverpod class.
```

docs/11-upgrading/01-upgrade-to-four.md
```diff
@@ -224,6 +224,7 @@ Copy the updated Dockerfile from the [4.0 framework template](https://github.com
 ## What's new in 4.0
 
 - **`serverpod start` TUI**: hot reload on save, **R** to hot restart, **M** to create a migration, **A** to apply migrations, **P** to apply a repair migration.
+- **Simplified server initialization**: the generated `Serverpod` class comes pre-wired with your project's `Protocol` and `Endpoints`, so `server.dart` creates the server with just `Serverpod(args)`. The classic `Serverpod(args, Protocol(), Endpoints())` form keeps working.
```
marcelomendoncasoares: Suggestion: Make this a little less verbose to match the list style.

```suggestion
- **Simplified server initialization** with the generated `Serverpod` class that pre-wires the `Protocol` and `Endpoints`; optional, `Serverpod(args, Protocol(), Endpoints())` still work.
```

marcelomendoncasoares (approved):
Suggestion: Only two minor suggestions, but I'll already leave it approved. Let's just wait a review of the @serverpod/docs group before merging.

docs/11-upgrading/01-upgrade-to-four.md
developerjamiu: Agreed on the length. One thing to add: the two forms need different imports. The generated class takes only `args` as a positional argument and does not re-export `Protocol` or `Endpoints`, so the old call only works if you keep the old imports.

```suggestion
- **Simplified server initialization** with the generated `Serverpod` class that pre-wires `Protocol` and `Endpoints`, so `server.dart` needs only `Serverpod(args)`. Projects that keep their existing imports can stay on `Serverpod(args, Protocol(), Endpoints())`.
```

docs/06-concepts/01-server-fundamentals/01-your-serverpod-project.md
FXschwartz: Good point, done! Applied the whole comment as one line since the suggestion range only covered the second line.

docs/11-upgrading/01-upgrade-to-four.md
FXschwartz: Done! Went with @developerjamiu version since it also captures that the two forms need different imports.

developerjamiu (approved):

PR #761, opened 2026-08-21
docs: Add web authentication with httpOnly cookies

FXschwartz:
Documents the `httpOnly` cookie web authentication that shipped in serverpod/serverpod#5555.

- New "Web setup" page under Authentication: server `authCookie` configuration, `allowedOrigins` and its CORS implications, client setup, per-token-manager behavior, the streaming contract, and local development notes.
- Points web users to cookie mode from the setup and token manager pages.
- Adds an "Authentication changes" section to the 4.0 upgrade guide: `?auth=` removal, the new sign-in policy, the TokenManager base class change for custom implementations, and stream closing on identity change.
- Renumbers custom overrides and legacy to make room.


docs/06-concepts/04-authentication/10-web-authentication.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+Everything else is unchanged: sign-in flows, the `client.auth` session manager, and endpoint calls work as on other platforms. Setting `cookieAuth` to `true` on a non-web platform throws, since those transports have no browser cookie jar.
+
+## How it works
+
+- With **server-side sessions**, the session token is delivered as an `httpOnly` cookie and never appears in the response body.
+- With **JWT**, the access token is kept in memory only, and the refresh token is delivered as an `httpOnly` cookie scoped to the refresh endpoint's path. On page load, the session is restored by refreshing from the cookie. Multiple tabs coordinate their refreshes through the browser's Web Locks API, so a shared refresh token is only rotated by one tab at a time.
+- **Signing out** clears the cookies and revokes the session on the server.
+- **Method streams** authenticate from the cookie at the WebSocket handshake. When the signed-in user changes (sign-in or sign-out), open method streams are closed gracefully — subscriptions receive `onDone` without an error — and new streams connect with the current identity. This applies on every platform, not only the web.
```
developerjamiu: Two em dashes in one sentence. The upgrade guide's version of this same bullet already uses parentheses, so the page can borrow its own wording.

```suggestion
- **Method streams** authenticate from the cookie at the WebSocket handshake. When the signed-in user changes (sign-in or sign-out), open method streams are closed gracefully (subscriptions receive `onDone` without an error) and new streams connect with the current identity. This applies on every platform, not only the web.
```

docs/11-upgrading/01-upgrade-to-four.md
```diff
@@ -221,6 +221,16 @@ Your production build needs to switch from `dart compile exe` to `dart build cli
...
 
+## Authentication changes
+
+4.0 changes a few authentication behaviors that can affect existing apps:
+
+- **The `?auth=` query parameter is no longer accepted.** Credentials never appear in URLs anymore: HTTP calls authenticate through the `Authorization` header (or an auth cookie on the web), and streaming connections authenticate in-band when the stream opens. Clients from before 4.0 that relied on the query parameter must be upgraded.
+- **Signing in on top of another account is rejected.** Issuing a token for a different user from an already-authenticated session throws a `SignInWhileAuthenticatedException` on every platform; users must sign out before switching accounts. Server code that mints tokens on behalf of another user (such as an admin flow) calls the token manager's `createToken` instead, which skips this policy and returns the secrets in the response body.
+- **Custom token managers extend a base class.** `TokenIssuer` and `TokenManager` are now base classes: implement `createToken` for the actual minting, and leave `issueToken` alone — it is non-virtual and applies the sign-in policy and cookie delivery for every token type.
```
developerjamiu: Em dash. Splitting into two sentences also lets the second one stand on its own, since the non-virtual behaviour is the part a custom implementer needs.

```suggestion
- **Custom token managers extend a base class.** `TokenIssuer` and `TokenManager` are now base classes: implement `createToken` for the actual minting, and leave `issueToken` alone. It is non-virtual and applies the sign-in policy and cookie delivery for every token type.
```

docs/06-concepts/04-authentication/10-web-authentication.md
```diff
@@ -0,0 +1,90 @@
...
+
+client = Client(serverUrl)
+  ..cookieAuth = kIsWeb
+  ..connectivityMonitor = FlutterConnectivityMonitor()
+  ..authSessionManager = FlutterAuthSessionManager();
+```
+
+Everything else is unchanged: sign-in flows, the `client.auth` session manager, and endpoint calls work as on other platforms. Setting `cookieAuth` to `true` on a non-web platform throws, since those transports have no browser cookie jar.
+
+## How it works
```
developerjamiu, line 67: Suggestion: this needs a `## Verify` between "Configure the client" and "How it works".

Both guide-shaped pages in this section already have one: `09-macos-authentication.md` right before this page, and `08-profile-photos.md`. Both use numbered steps where each step has an outcome the reader can see.

It matters more here than on either of those. Success is invisible by design: after setting `cookieAuth = kIsWeb` nothing on screen changes, and the two likely mistakes both look identical to an ordinary sign-in failure, an origin missing from `allowedOrigins` and `secure: true` left on for localhost.

Something like: sign in, confirm the auth cookie appears under Application in devtools, confirm nothing lands in `localStorage`, then reload and confirm the session survives. You can check the exact devtools wording.

docs/06-concepts/04-authentication/10-web-authentication.md
developerjamiu: Applied in 9e19865.

docs/11-upgrading/01-upgrade-to-four.md
developerjamiu: Applied in 9e19865.

docs/06-concepts/04-authentication/10-web-authentication.md
developerjamiu: Added in 9e19865, with steps grounded in the framework's cookie tests: the cookie name per token manager, the HttpOnly flag, the empty Local Storage check, and the reload.

developerjamiu: Reworded step 2 in e64d5cb: it now checks the Set-Cookie header on the sign-in response in the Network tab, which is what the framework tests assert, instead of the cookies panel, which is not reliable when the server is a different origin.

developerjamiu: One more correction on step 2: in JWT mode only the refresh token moves to the cookie; the sign-in body still carries the short-lived access token (the test asserts body token isNotEmpty with refreshToken null). The step now states that, so a reader inspecting the body is not surprised by the access token.

FXschwartz:
@developerjamiu Thank you for filling in the blanks on this one, really appreciate they help! All of your additions look great and make sense.

CC @marcelomendoncasoares

developerjamiu (approved):
LGTM

PR #767, opened 2026-08-27
docs: Clarify how endpoint names map to client properties

Benzzick:
## Summary

Adds a tip explaining how Serverpod endpoint class names map to
properties on the generated client.

For example:

- `RecipeEndpoint`
- `client.recipe`

This helps clarify why endpoint methods are called through
`client.recipe.generateRecipe(...)` after code generation.

### DIFF
--- a/versioned_docs/version-3.4.0/01-get-started/01-creating-endpoints.md
+++ b/versioned_docs/version-3.4.0/01-get-started/01-creating-endpoints.md
@@ -106,6 +106,10 @@ $ serverpod generate
 
 `serverpod generate` will create bindings for the endpoint and register them in the server's `generated/protocol.dart` file. It will also generate the required client code so that you can call your new `generateRecipe` method from your app.
 
+:::tip Endpoint names become client properties
+In the next section you'll call `client.recipe.generateRecipe(...)` — not `client.recipeEndpoint.generateRecipe(...)` or `client.generateRecipe(...)`. That's not a typo: Serverpod derives the property name from your endpoint's class name by dropping the `Endpoint` suffix and lowercasing the first letter, so `RecipeEndpoint` becomes `client.recipe`. Every method you add inside the class shows up under that property, e.g. `client.recipe.generateRecipe(...)`. This mapping is generated automatically by `serverpod generate` above — there's nothing to configure by hand, but it's worth knowing the rule so the client code doesn't feel like it's coming out of nowhere.
+:::
+
 :::note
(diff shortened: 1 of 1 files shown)


versioned_docs/version-3.4.0/01-get-started/01-creating-endpoints.md
```diff
@@ -106,6 +106,10 @@ $ serverpod generate
 
 `serverpod generate` will create bindings for the endpoint and register them in the server's `generated/protocol.dart` file. It will also generate the required client code so that you can call your new `generateRecipe` method from your app.
 
+:::tip Endpoint names become client properties
+In the next section you'll call `client.recipe.generateRecipe(...)` — not `client.recipeEndpoint.generateRecipe(...)` or `client.generateRecipe(...)`. That's not a typo: Serverpod derives the property name from your endpoint's class name by dropping the `Endpoint` suffix and lowercasing the first letter, so `RecipeEndpoint` becomes `client.recipe`. Every method you add inside the class shows up under that property, e.g. `client.recipe.generateRecipe(...)`. This mapping is generated automatically by `serverpod generate` above — there's nothing to configure by hand, but it's worth knowing the rule so the client code doesn't feel like it's coming out of nowhere.
+:::
```
developerjamiu, line 111: The tip covers a real gap: the 4.0 version of this page explains this mapping and the 3.4 page doesn't. Two things about the form though. It sits directly above an existing `:::note`, so the section runs prose, tip, note, heading, and callouts stacked like that compete rather than stand out. It also introduces the only em dash on this page.

The 4.0 page carries the same fact as one clause in a paragraph, which reads lighter here too:

```suggestion
`serverpod generate` will create bindings for the endpoint and register them in the server's `generated/protocol.dart` file. It will also generate the required client code so that you can call your new `generateRecipe` method from your app. The class name's `Endpoint` suffix is dropped and the first letter lowercased, so `RecipeEndpoint` is called via `client.recipe`, and every method in the class appears under that property.
```

versioned_docs/version-3.4.0/01-get-started/01-creating-endpoints.md
```diff
@@ -241,4 +245,4 @@ Try out the app by clicking the button to get a new recipe. The app will call th
 
 ## Next steps
```
developerjamiu, line 247: This drops the trailing newline on a line the change doesn't otherwise touch. Restore it.

developerjamiu (approved):

PR #771, opened 2026-08-31
docs(cloud): Add a storage concept page

Isakdl:
Adds `cloud_docs/concepts/storage.md`, a concept page for Serverpod Cloud storage.

It covers the two storages every project starts with, choosing access, using storage from server code, managing storages and their files from the CLI and the console, and what happens at a plan limit.

The blockers listed on this PR are resolved. The provider API shipped in Serverpod 4.0.0, projects are created with both default storages, and the console Storage tab is live and no longer gated. The page is updated against Serverpod 4.0.0 and Cloud CLI 1.0.0:

- Server-side wiring uses `ServerpodCloudProvider.private()` and `.public()`, which `serverpod create` already adds to new projects, and the `session.storage` examples use the shipped method names.
- Adds the `serverpod cloud storage` commands, which the first draft did not cover, including creating and deleting a storage.
- Keeps custom storages reachable from server code, and says plainly that Cloud connects the two defaults for you while a storage you add has no ready-made helper, with a pointer to how storage providers are registered.
- Limits are scoped to the plans that set caps, and describe what a lockout affects and how access comes back.
- Treats `private` and `public` as storage ids named after their access rather than as fixed categories.

The console workflow, its labels, the storage id rules, and the type-the-storage-id delete confirmation were all checked against source.

### DIFF
--- a/cloud_docs/concepts/storage.md
+++ b/cloud_docs/concepts/storage.md
@@ -1,0 +1,141 @@
+---
+sidebar_position: 9
+description: Every Serverpod Cloud project gets file storage for avatars, documents, and other user files, read and written through the standard Serverpod storage API.
+---
+
+# Storage
+
+Avatars, invoice PDFs, video attachments, and generated exports do not belong in your database. Serverpod Cloud keeps them as files instead. Your app uploads a file straight to storage rather than through an endpoint method, which keeps large files out of your API calls. Every new Serverpod Cloud project starts with two storages, `private` and `public`, matching the two the Serverpod framework configures by default.
+
+## Choose between private and public
+
+`private` is the default. Only your server can read the files in it. When your app needs a file, your server hands out a link that works for a short time.
+
+`public` serves its files to anyone who has the URL. Use it for content you would put on a website, such as profile pictures, product images, and downloadable assets.
+
+You choose access when you create a storage, and it cannot be changed afterwards. Pick `private` when you are unsure. You can add a public storage later and copy the files you want to expose into it.
+
+## Use a storage from your server
+
+Register each storage before you start the server. The storage id is the name you gave it.
+
+<!-- TODO: ServerpodCloudStorage has not landed in the framework yet. Replace this
+     block with the real class, its package, and its import once it ships. -->
+
+```dart title="server.dart"
+void run(List<String> args) async {
+  final pod = Serverpod(args, Protocol(), Endpoints());
+
+  pod.addCloudStorage(await ServerpodCloudStorage.create(storageId: 'private'));
+  pod.addCloudStorage(await ServerpodCloudStorage.create(storageId: 'public'));
+
+  await pod.start();
+}
+```
+
+`ServerpodCloudStorage.create` is asynchronous, so call it before `pod.start()`. Projects created from the Serverpod template already contain these two lines.
+
+After that, reach the storage through `session.storage`, the same API every Serverpod storage uses.
+
+Write a file from your server:
+
+```dart
+await session.storage.storeFile(
+  storageId: 'private',
+  path: 'invoices/2026/$invoiceId.pdf',
+  byteData: pdfBytes,
+);
+```
+
+Read it back:
+
+```dart
+final pdfBytes = await session.storage.retrieveFile(
+  storageId: 'private',
+  path: 'invoices/2026/$invoiceId.pdf',
+);
+```
+
+Get the URL of a file in a public storage:
+
+```dart
+final url = await session.storage.getPublicUrl(
+  storageId: 'public',
+  path: 'avatars/$userId.png',
+);
+```
+
+To upload from your app, your server creates an upload description and your app sends the file with it. See [File uploads](https://docs.serverpod.dev/concepts/endpoints-and-apis/file-uploads) for the full flow, including the Flutter side.
+
+```dart
+class ProfileEndpoint extends Endpoint {
+  Future<String?> getAvatarUploadDescription(Session session, String path) {
+    return session.storage.createDirectFileUploadDescription(
+      storageId: 'public',
+      path: path,
+    );
+  }
+
+  Future<bool> verifyAvatarUpload(Session session, String path) {
+    return session.storage.verifyDirectFileUpload(
+      storageId: 'public',
+      path: path,
+    );
+  }
+}
+```
+
+## Add a storage
+
+Add a storage when you want files kept apart from the two defaults, such as one per tenant or one for exports you purge on a schedule.
+
+1. Open your project in the Cloud console and select the **Storage** tab.
+2. Select **Create storage**.
+3. Enter a storage id. This is the name your code passes as `storageId`. Use lowercase letters, digits, and dashes.
+4. Choose **Private** or **Public**.
+5. Select **Create storage**.
+
+The storage shows as `Creating` for a few seconds, then as `Private` or `Public`.
+
+Deploy your project again so the new storage reaches your running server:
+
+```bash
+scloud deploy
+```
+
+Then register it the same way as the default storages:
+
+<!-- TODO: same placeholder class as above. -->
+
+```dart title="server.dart"
+pod.addCloudStorage(await ServerpodCloudStorage.create(storageId: 'user-uploads'));
+```
+
+## Browse files in the console
+
+Select a storage in the **Storage** tab to open its file browser.
+
+A storage holds a flat list of files, and the console reads `/` in a path as a folder separator. A file stored at `avatars/2026/user-42.png` appears under `avatars`, then `2026`. The path you write is the only structure you get, so decide on a path scheme before you store many files. Use **Filter by name** to narrow a long list.
+
+To add files, select **Upload files**, or drag them onto the browser. Select **Upload folder** to upload a whole folder and keep its structure.
+
+Each file row has **Download** and **Delete**.
+
+## Delete a storage
+
+Open the row menu for the storage and select **Delete storage**. Type the storage id to confirm.
+
+Deleting a storage deletes every file in it. The files cannot be recovered. Remove the storage from `server.dart` and deploy again, or calls to that storage id will fail.
+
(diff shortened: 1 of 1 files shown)


developerjamiu:
Reworked against Serverpod 4.0.0 and Cloud CLI 1.0.0.

- The placeholder class is now `ServerpodCloudProvider.private()` and `.public()`, and three `session.storage` method names are corrected.
- Added the `serverpod cloud storage` commands.
- Limits now scope the lockout to the plans that set caps.

Your blockers are resolved: the provider API shipped, projects get both default storages, and the Storage tab is live and ungated.

I first cut the custom-storage registration step, thinking server code could reach only `private` and `public`. That was wrong, so it is back, worded to say there is no ready-made helper for a custom storage id.

cloud_docs/concepts/storage.md
```diff
@@ -0,0 +1,127 @@
...
+
+A storage is either private or public, and the two you start with are named after their access.
+
+Files in a private storage are never served publicly. Your server reads and writes them, and you can reach them yourself from the CLI and the console. Use it for anything belonging to a single user or anything you would not publish. To let an app download one, your server hands out a link that works for a short time with `session.storage.temporaryDownloadUrl`.
+
+Files in a public storage are served to anyone who has the URL. Use it for content you would put on a website, such as profile images, product photos, and downloadable assets.
+
+Access is fixed when a storage is created and cannot be changed afterwards. Choose private when you are unsure. You can create a public storage later and copy the files you want to expose into it.
+
+## Use storage from your server
```
SandPod, line 21: Question: Do we need this copied over in the cloud documentation?

The risk is that we forget to update this here if we ever change it in Serverpod. Perhaps it is better to link to that documentation instead? And focus more on the Serveprod Cloud aspects here?

cloud_docs/concepts/storage.md
```diff
@@ -0,0 +1,127 @@
...
+  storageId: 'public',
+  path: 'avatars/$userId.png',
+);
+```
+
+To upload from your app instead, your server creates an upload description and your app sends the file with it. See [File uploads](/concepts/endpoints-and-apis/file-uploads) for the full flow, including the Flutter side.
+
+Cloud connects `private` and `public` for you. A storage you add yourself has no ready-made helper, so you register it in `server.dart` the same way you would any other storage provider. See [Configure a storage provider](/concepts/endpoints-and-apis/file-uploads#configure-a-storage-provider) for how registration works.
+
+## Manage your storages
```
SandPod: Question: General question for the docs - How do we view giving instructions in the terminal vs the console? Should we have both?

cloud_docs/concepts/storage.md
```diff
@@ -0,0 +1,127 @@
...
+
+A storage id uses lowercase letters, digits, and dashes, starts and ends with a letter or a digit, and is at most 63 characters. New storages are private unless you pass `--access public`.
+
+Delete a storage and everything in it:
+
+```bash
+serverpod cloud storage delete exports
+```
+
+The command asks you to confirm. The files cannot be recovered, and code that still writes to that storage id fails afterwards.
```
SandPod: Question: This scentence reads a bit off. Perhaps we can improve it by talking about the files now being recorable before the `bash` command and then only mention that you need to confirm deletion afterwards and that writes will fail for that id?

SandPod (commented):
Have a couple of questions for you before I approve.

I assume that all has been tested with the latest version of the CLI/Console.

cloud_docs/concepts/storage.md
```diff
@@ -0,0 +1,127 @@
...
+description: Serverpod Cloud gives every project file storage. Use it from your server through session.storage, and manage storages and their files from the CLI or the console.
+---
+
+# Storage
+
+Avatars, invoice PDFs, and generated exports do not belong in your database. Serverpod Cloud keeps them as files instead. Every new project starts with two storages, `private` and `public`, matching the two the Serverpod framework configures by default, and you can add more.
+
+## Choose access for a storage
+
+A storage is either private or public, and the two you start with are named after their access.
```
Isakdl: Comment: This isn't true, the "storage id" can be set to anything on a storage bucket, by default we create 2 storage buckets with the ids `private` and `public`.

cloud_docs/concepts/storage.md
```diff
@@ -0,0 +1,127 @@
...
+
+Files in a private storage are never served publicly. Your server reads and writes them, and you can reach them yourself from the CLI and the console. Use it for anything belonging to a single user or anything you would not publish. To let an app download one, your server hands out a link that works for a short time with `session.storage.temporaryDownloadUrl`.
```
Isakdl: Comment: I believe private storage can still serve files publicly if the URL is signed, but they are not accessible unless you have the url.

cloud_docs/concepts/storage.md
```diff
@@ -0,0 +1,127 @@
...
+```dart
+final url = await session.storage.publicDownloadUrl(
+  storageId: 'public',
+  path: 'avatars/$userId.png',
+);
+```
+
+To upload from your app instead, your server creates an upload description and your app sends the file with it. See [File uploads](/concepts/endpoints-and-apis/file-uploads) for the full flow, including the Flutter side.
+
+Cloud connects `private` and `public` for you. A storage you add yourself has no ready-made helper, so you register it in `server.dart` the same way you would any other storage provider. See [Configure a storage provider](/concepts/endpoints-and-apis/file-uploads#configure-a-storage-provider) for how registration works.
```
Isakdl: Suggestion: Add an example of what this looks like

cloud_docs/concepts/storage.md
```diff
@@ -0,0 +1,127 @@
...
+
+In the console, select a storage in the **Storage** tab to open its file browser. A storage holds a flat list of files, and the console reads `/` in a path as a folder separator. A file stored at `avatars/2026/user-42.png` appears under `avatars`, then `2026`. The path you write is the only structure you get, so decide on a path scheme before you store many files. Use **Filter by name** to narrow a long list.
+
+To add files, select **Upload files**, or drag them onto the browser. Select **Upload folder** to upload a whole folder and keep its structure. Each file row has a menu with **Download** and **Delete**.
+
+## Limits
+
+- **Storages per project.** Your plan sets how many storages a project can have.
+- **Metered usage.** Serverpod Cloud meters three things and bills them by usage: the data you store, the data read out of your storages, and the operations performed on them.
+- **Caps on some plans.** Plans that set caps lock a project out when it goes over one. Every storage in the project becomes unreadable and unwritable, and public URLs stop working.
```
Isakdl: Comment: The user cannot set a cap on usage

Isakdl (commented):
Looking good, but added some clarification comments on some section that I think needs a rework.

cloud_docs/concepts/storage.md
developerjamiu: We should do both. We write the steps for the terminal and mention that the console does the same thing. Writing the steps twice means the console version goes out of date when the console changes. Full console instructions belong in the console documentation.

cloud_docs/concepts/storage.md
developerjamiu: We should link to the framework docs instead. Will do that.

cloud_docs/concepts/storage.md
developerjamiu: Moved up

cloud_docs/concepts/storage.md
developerjamiu: I just did some findings.

There is no example to show yet. Connecting a storage goes through `ServerpodCloudProvider`, which only has `private()` and `public()`, so `session.storage` throws for any other id. The generic function behind those takes a storage id and would work, but it is not exported. I can add the example as soon as there is a public way to call it, so is a method like `forStorage(id)` helper planned? @Isakdl

cloud_docs/concepts/storage.md
developerjamiu: Applied. The page now says any storage id can be private or public, and that the two you start with happen to be named after their access.

cloud_docs/concepts/storage.md
developerjamiu: Applied

cloud_docs/concepts/storage.md
developerjamiu: Applied

cloud_docs/concepts/storage.md
Isakdl: what? do we not support generic ids?

Isakdl: It looks like we do support this? https://github.com/serverpod/serverpod/blob/c804637f81d8b1a6c4f3fa16fdb6434a501e40aa/integrations/serverpod_cloud_storage/lib/src/serverpod_cloud_storage.dart#L44

This is a public method, to set your own id. One annoying thing with that interface is that you have to supply Platform.environment manually, but if that is the interface for now I would document this.

Isakdl: Oh I understand the issue now, why this method cannot be called, but this is a big issue. We are selling this as a premium feature in the growth tier.

developerjamiu: Added, now that ServerpodCloudProvider.custom is in 4.0.2 (serverpod/serverpod#5738). The API itself goes on the framework file-uploads page in a separate PR.

cloud_docs/concepts/storage.md
SandPod: Framework code is still part of the docs, shouldn't this be a "guide" in that case? Or do we want it in the docs?

SandPod (approved):
Looks good from my perspective.

But still unsure about the code examples from the Framework. Easy that this gets outdated. Perhaps this is better suited for a guide?

developerjamiu:
The code left here is Cloud's own wiring, not the framework API. The API lives on the file-uploads page (#857), and the CDN and passwords pages carry the same kind of snippet.

PR #774, opened 2026-09-01
docs: Add cloud secret redirect and status/context intros

developerjamiu:
Follow-up to #772, which removed the `scloud secret` reference pages after CLI 0.38.0 merged the command into `scloud variable --secret`.

- Redirects `/cloud/reference/cli/commands/secret` to the variable page. Released CLIs up to 0.37.x print the old URL in their help output, so the link keeps working for users on those versions. The redirect could only land once #772 deleted the page, since the plugin rejects redirects whose source path still exists.
- Writes the curated intro for the new `scloud status` page, which shipped with an empty scaffold and rendered without a heading or title. The intro defines podlets in passing, since the generated help uses the term and no other page does.
- Fills in the `scloud context` intro, which had the same gap since the page was created, and points readers choosing between context and per-directory linking at `scloud project link`.


Zfinix (approved):
LGTM

PR #775, opened 2026-09-01
docs: Document the SPA and static cache control environment variables

developerjamiu:
Companion to #756. The web server ships three cache-control environment variables built on the same mechanism; #756 documents the `FlutterRoute` one, and this adds the other two on their own pages:

- `SERVERPOD_WEB_SERVER_STATIC_CACHE_CONTROL` on the Static files page, next to the cache-control helpers it substitutes for.
- `SERVERPOD_WEB_SERVER_SPA_CACHE_CONTROL` on the Single-page apps page, including the two SpaRoute-specific rules: the fallback file is always served without a `Cache-Control` header, and the static variable does not apply to `SpaRoute`.

In all three cases an explicit `cacheControlFactory` takes precedence over the environment variable.


docs/06-concepts/05-web-server/05-static-files.md
```diff
@@ -52,6 +52,14 @@ pod.webServer.addRoute(
 );
 ```
 
+To set the header per environment instead of in code, set `SERVERPOD_WEB_SERVER_STATIC_CACHE_CONTROL` to the `Cache-Control` value the route should send:
```
SandPod, line 55: Question - Have we listed these on the ENV page? What is the strategy around that currently?

developerjamiu: Not yet (thanks for the catch). The reference only listed variables that also have a config-file option, and these three are env-only. I'll add all three (including #756's FlutterRoute one) to the table with links to their pages. Going forward the rule is: every SERVERPOD_* variable gets a row in the reference, and the full explanation lives on the feature's own page.

SandPod: Sounds good! 👍

SandPod (approved):
LGTM 👍

PR #779, opened 2026-09-03
docs: Add scalability and optimization guidance

### LINKED ISSUE #1718: Documentation page on scalability and optimization
I looked at the new docs, and they do cover elastic scaling of Serverpod serverless instances, but there are a lot more issues that should be covered in a separate server scalability page, for users whose Serverpod-powered apps may suddenly go viral, and have to scale to millions of daily active users. I will list some of the issues I have come across here.

* Offload any computationally-intensive operations in the server into a new isolate (via `Isolate.run` or similar), so that you don't hold up the main thread, since there is only one thread responding to requests per Serverpod instance. (viz #1423)
* Postgres query optimization:
  * Ensure that indexes exist, not just for terms in `WHERE` clauses, but also for terms in `ORDER BY` clauses (if you don't have an index for all terms in an `ORDER BY` expression, then the whole query result has to be materialized before an `OFFSET...LIMIT` restriction can be applied, so potentially a lot more work has to be done for paged queries without these indexes).
  * Avoid using `OR` and (equivalently) `IN SET` queries (if you care about high scalability), especially when you have multiple `OR` clauses linked by `AND` operators, since the disjunctive normal form of these expressions will explode multiplicatively.
* Postgres server scalability:
  * How to specify the number of cores to assign to a Postgres server instance, and how to add cores if necessary.
  * How to set up read-only Postgres replicas, to scale up the read-only query throughput. (This probably requires Terraform script support, and additionally, Serverpod may need some modifications, so that for example, in addition to `MyTable.db`, there is a `MyTable.dbro`, for read-only queries, so that they can be load balanced across replicas. Or potentially Serverpod can automatically determine which queries are read-only, and submit to the replica load balancer. Or maybe a Postgres replica load balancer already does all this automatically, and sends update requests just to the write (master) replica -- I don't know.)
  * How to configure Postgres tuning params for optimal performance on cloud instances, relative to expected workload. I haven't looked into this in any detail yet, but here is an example set of tuning parameters that I found:
    ```
    shared_buffers='36GB'
    max_connections=1000
    checkpoint_completion_target='0.93'
    checkpoint_timeout='30min'
    max_wal_size='300GB'
    effective_cache_size='100GB'
    random_page_cost=1
    seq_page_cost=1
    effective_io_concurrency=200
    ```
* Serverpod server scalability:
  *  Don't use a streaming connection if you think you will have more than 64k concurrent connections per Serverpod node (since this is the limit on the number of ports per IP address). Instead, use HTTPS requests from client to server, and push notifications (via Firebase Cloud Messaging or similar) from server to client.
  * Use an `AuthenticationHandler` that does not hit the database to fetch the `UserInfo` object on every request (this needs to be written: #1720).

abdulawalarif:
## Summary
- Adds an Operations concept page on scaling Serverpod beyond a single process: server roles, `Isolate.run` for CPU work, Postgres indexes and query shape, connection pool sizing, Redis for shared cache and events, JWT auth, and streaming tradeoffs.
- States current limits honestly (no first-class read replicas or Serverpod-owned Postgres tuning), and points to database interceptors or an external proxy where needed.
- Links the page from the Operations overview and Hosting elsewhere (server roles).

Closes [serverpod/serverpod#1718](https://github.com/serverpod/serverpod/issues/1718).

## Test plan
- [ ] Open Next → Concepts → Operations → Scalability.
- [ ] Confirm the page appears in the Operations sidebar after Exception monitoring.
- [ ] Follow inbound links from Operations overview and Hosting elsewhere → Server roles.
- [ ] Spot-check outbound links: server roles, indexing, pagination, connection pool, database interceptors, JWT token manager, caching, server events, streaming, configuration reference.
- [ ] Skim for em dashes and marketing wording against STYLE_GUIDE.md.

### DIFF
--- a/docs/06-concepts/07-operations/06-scalability.md
+++ b/docs/06-concepts/07-operations/06-scalability.md
@@ -1,0 +1,131 @@
+---
+description: Scale Serverpod beyond a single process with server roles, isolates, Postgres indexes and pools, Redis, JWT auth, and when short RPCs beat long-lived streams.
+---
+
+# Scalability
+
+When traffic jumps, the first bottlenecks are usually a blocked Dart event loop, an undersized database pool across many nodes, or shared state that only lives in one process. This page covers how Serverpod scales out request-serving instances, and what to change so CPU work, Postgres, auth, caching, and long-lived connections keep up.
+
+Elastic scale-out of request handlers is already covered by [server roles](../server-fundamentals/running-your-server#choose-a-server-role). The sections below cover what still bites at high daily active users. Built-in read replicas and a Serverpod-owned Postgres tuning recipe are not available today; use cloud provider guides and the escape hatches noted here.
+
+## How Serverpod scales
+
+Each Serverpod process runs on one Dart isolate and one event loop. CPU-heavy work on that isolate delays request handling for every concurrent call on the same process.
+
+HTTPS APIs with JWT auth are largely stateless and scale behind a load balancer. Features that keep in-process state (local caches, local server events, long-lived streams without a shared bus) need either sticky routing, Redis, or a redesign so state lives outside the process.
+
+For serverless or auto-scaled fleets, run request nodes in the `serverless` role (or `monolith` if you accept in-process maintenance), and run `maintenance` on a schedule for future calls and health metric collection. See [Hosting elsewhere](../../deployments/custom-hosting/hosting-elsewhere#server-roles) and [Scheduling overview](../scheduling/overview).
+
+## Split request and maintenance work
+
+Use roles so request capacity and background work do not share the same scaling curve:
+
+| Role | Serves requests | Runs future calls and health metrics |
+| --- | --- | --- |
+| `monolith` | Yes | Yes (continuous) |
+| `serverless` | Yes | No |
+| `maintenance` | No | Once, then exit |
+
+```bash
+dart run bin/main.dart --mode production --role serverless
+```
+
+Schedule a separate process in the `maintenance` role (for example once per minute) when request nodes run as `serverless`. On pure request nodes you can also set `futureCallExecutionEnabled` to `false`; see [Future call configuration](../scheduling/configuration).
+
+## Offload CPU work to isolates
+
+Image decoding, password hashing, and other CPU-bound work should not run on the request isolate. Use Dart's `Isolate.run` with plain data only:
+
+```dart
+Future<List<int>> buildThumbnail(List<int> bytes) async {
+  return Isolate.run(() {
+    // CPU-heavy decode, resize, and encode.
+    // Pass only serializable data in and out.
+    return processImage(bytes);
+  });
+}
+```
+
+Do not pass a `Session`, database connection, or Redis client into the isolate. Do I/O on the main isolate after the isolate returns.
+
+There is no first-party Serverpod helper for this pattern. For sustained background workloads, raise `futureCall.concurrencyLimit` carefully (default `1`), or plan for a dedicated worker model as your product grows.
+
+## Optimize Postgres queries
+
+Declare indexes in your models for columns you filter and sort on. Without an index that covers an `ORDER BY`, Postgres may sort the full result set before applying `OFFSET` / `LIMIT`. See [Indexing](../data-and-the-database/database/indexing).
+
+Prefer [cursor-based pagination](../data-and-the-database/database/pagination#cursor-based-pagination) over deep `offset` pages on large tables: filter with `where: (t) => t.id > lastId`, `orderBy` on `id`, and a `limit`.
+
+At high scale, treat wide `OR` / `inSet` filters carefully, especially expressions like `(A OR B) AND (C OR D)`. Those shapes can explode into many indexed paths. Prefer a denormalized flag, a `UNION` of simpler indexed queries, or a narrower filter design when those queries dominate load.
+
+## Size the connection pool
+
+Each Serverpod process opens its own pool. The default `database.maxConnectionCount` is `10` (override with `SERVERPOD_DATABASE_MAX_CONNECTION_COUNT`). Across a fleet:
+
+`(number of nodes × maxConnectionCount) + headroom ≤ Postgres max_connections`
+
+Leave headroom for Insights, maintenance jobs, migrations, and admin tools. Unlimited pools (`0` or a negative value) are easy to misconfigure under auto-scaling. See [Database connection](../data-and-the-database/database/connection#configure-connection-pool-size).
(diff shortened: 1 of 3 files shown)


docs/06-concepts/07-operations/06-scalability.md
```diff
@@ -0,0 +1,131 @@
...
+
+Modern projects use the [JWT token manager](../authentication/token-managers/jwt-token-manager). Access-token verification is local and does not query the database on every request. Refresh still hits the database; that path is far less frequent than ordinary API calls.
+
+[Server-side sessions](../authentication/token-managers/managing-tokens) are the opposite tradeoff: a database lookup on validation, with immediate revocation. Prefer JWT when request volume is the constraint. Legacy auth key handlers that load a row per request do not belong on a high-QPS fleet.
+
+## Share state across instances with Redis
+
+Local caches (`session.caches.local` and `localPrio`) stay on one process. For values every instance must see, use the [global cache](../endpoints-and-apis/caching#the-global-cache-and-redis) with Redis enabled.
+
+Cross-instance [server events](../endpoints-and-apis/server-events) need Redis as well. `MessageScope.global` requires Redis; without it, messaging stays local to the process that posted the event.
```
developerjamiu: `MessageScope.global` does not stay local without Redis, it throws (`StateError('Redis needs to be enabled to use this method')` in message_central.dart). The local fallback is `MessageScope.auto`'s behavior.

```suggestion
Cross-instance [server events](../endpoints-and-apis/server-events) need Redis as well. `MessageScope.global` throws a `StateError` without Redis, and the default `MessageScope.auto` falls back to local delivery, so without Redis events silently stop reaching other instances.
```

docs/06-concepts/07-operations/06-scalability.md
```diff
@@ -0,0 +1,131 @@
...
+
+At high scale, treat wide `OR` / `inSet` filters carefully, especially expressions like `(A OR B) AND (C OR D)`. Those shapes can explode into many indexed paths. Prefer a denormalized flag, a `UNION` of simpler indexed queries, or a narrower filter design when those queries dominate load.
+
+## Size the connection pool
+
+Each Serverpod process opens its own pool. The default `database.maxConnectionCount` is `10` (override with `SERVERPOD_DATABASE_MAX_CONNECTION_COUNT`). Across a fleet:
+
+`(number of nodes × maxConnectionCount) + headroom ≤ Postgres max_connections`
+
+Leave headroom for Insights, maintenance jobs, migrations, and admin tools. Unlimited pools (`0` or a negative value) are easy to misconfigure under auto-scaling. See [Database connection](../data-and-the-database/database/connection#configure-connection-pool-size).
```
developerjamiu: The framework documents unlimited as null or negative, not `0` ("a negative or null value removes the limit" in the shipped config templates; the config doc comment says null means unlimited).

```suggestion
Leave headroom for Insights, maintenance jobs, migrations, and admin tools. Unlimited pools (a negative value, or omitting the key) are easy to misconfigure under auto-scaling. See [Database connection](../data-and-the-database/database/connection#configure-connection-pool-size).
```

docs/06-concepts/07-operations/06-scalability.md
```diff
@@ -0,0 +1,131 @@
...
+
+| Setting | Default | Why it matters at scale |
+| --- | --- | --- |
+| `role` / `SERVERPOD_SERVER_ROLE` | `monolith` | Split request nodes from maintenance work. |
+| `database.maxConnectionCount` | `10` | Pool size times node count must fit Postgres. |
+| `redis.enabled` | `false` in templates | Required for shared cache and global events. |
+| `maxRequestSize` | `524288` | Large uploads increase memory pressure. |
+| `websocketPingInterval` | `30` (seconds) | Keepalive cost under many open streams. |
+| `futureCall.concurrencyLimit` | `1` | Caps background CPU and database load. |
+| `sessionLogs.persistentEnabled` | mode-dependent | Extra database writes per request when on. |
```
developerjamiu: The default is database-dependent, not mode-dependent: `SessionLogConfig.buildDefault` sets `persistentEnabled: databaseEnabled`. The run mode only affects console logging.

```suggestion
| `sessionLogs.persistentEnabled` | on when a database is configured | Extra database writes per request when on. |
```

docs/06-concepts/07-operations/06-scalability.md
```diff
@@ -0,0 +1,131 @@
...
+| `healthCheckInterval` | about 1 minute | Metrics write load; aggressive liveness elsewhere can cascade restarts. |
```
developerjamiu: It is exactly `Duration(minutes: 1)`.

```suggestion
| `healthCheckInterval` | 1 minute | Metrics write load; aggressive liveness elsewhere can cascade restarts. |
```

docs/06-concepts/07-operations/06-scalability.md
```diff
@@ -0,0 +1,131 @@
...
+
+Keep `websocketPingInterval` (default 30 seconds) in mind under high connection counts: each open socket pays keepalive cost.
+
+## Configuration cheat sheet
+
+| Setting | Default | Why it matters at scale |
+| --- | --- | --- |
+| `role` / `SERVERPOD_SERVER_ROLE` | `monolith` | Split request nodes from maintenance work. |
+| `database.maxConnectionCount` | `10` | Pool size times node count must fit Postgres. |
+| `redis.enabled` | `false` in templates | Required for shared cache and global events. |
```
developerjamiu: It is the config default, not only a template choice; the configuration reference lists the default as false.

```suggestion
| `redis.enabled` | `false` | Required for shared cache and global events. |
```

abdulawalarif:
Addressed the suggestions, Jamiu. 
Please take another look.

docs/06-concepts/07-operations/06-scalability.md
```diff
@@ -0,0 +1,131 @@
+---
+description: Scale Serverpod beyond a single process with server roles, isolates, Postgres indexes and pools, Redis, JWT auth, and when short RPCs beat long-lived streams.
```
developerjamiu: Sibling pages lead the frontmatter description with a noun phrase (122 of 126 in Concepts). Noun-led form:

```suggestion
description: "Scaling guidance for Serverpod: server roles, isolates for CPU work, Postgres indexes and pools, Redis, JWT auth, and when short RPCs beat long-lived streams."
```

abdulawalarif:
Hey Jamiu, updated the frontmatter description to the noun-led

developerjamiu (approved):

PR #791, opened 2026-09-08
docs: Fix broken framework links in Cloud documentation

developerjamiu:
Eight links from four Cloud pages pointed at framework paths that only exist in the 4.0 prerelease. Readers on the released version hit 404s while following the deploy, logging, caching, and web-hosting paths.

They were all written as absolute `https://docs.serverpod.dev/...` URLs. Docusaurus treats those as external and skips them, so the broken-link check never saw them. Each one now points at the equivalent page in the served version, written as a root-relative path like the rest of our cloud-to-framework links, so the build checks it from here on.

### DIFF
--- a/cloud_docs/concepts/cdn.md
+++ b/cloud_docs/concepts/cdn.md
@@ -62,7 +62,7 @@ cacheControlFactory: StaticRoute.public(maxAge: const Duration(minutes: 1)),
 Flutter's build uses fixed file names with no content hashes, so a long browser `max-age` can leave a visitor on an old build after you deploy. Keep the browser lifetime short and let the deploy purge handle the edge.
 :::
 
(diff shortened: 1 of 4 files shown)


SandPod (commented):
Great! Can you apply these to all older versions of the documentation where this applies as well?

developerjamiu:
The Cloud docs are a single unversioned instance, so these four pages have no older copies to update. I checked all 19 versioned framework snapshots too, and none contains a `docs.serverpod.dev` link in any form.

On older-version readers: an unversioned Cloud page can't know which Serverpod version someone is on, so these links resolve to the current stable framework docs for everyone. That's the same behavior as every other cloud-to-framework link we have.

SandPod (approved):
LGTM! 👍

PR #805, opened 2026-09-08
docs: Fix migration tag example, cookie-auth example, and config reference gaps

### LINKED ISSUE #792: docs: Upgrade guide migration command uses a tag the CLI rejects
The 4.0 upgrade guide tells users to run `serverpod create-migration --tag "upgrade-4.0"`. Tag names must match `^[a-z0-9]+([-][a-z0-9]+)*$`, so the dot fails with `"Tag names can only contain lowercase letters, numbers, and dashes."` Change the example to `--tag "upgrade-4-0"` and state the tag rule on the migrations page and the create-migration reference intro.

Files: docs/11-upgrading/01-upgrade-to-four.md:131, docs/06-concepts/03-data-and-the-database/02-database/13-migrations.md, docs/06-concepts/cli/commands/create-migration/_create-migration.md Source: tools/serverpod_cli/lib/src/util/string_validators.dart:18-20, lib/src/commands/create_migration.dart:51-57

### LINKED ISSUE #793: docs: Cookie-auth config example silently leaves cookie auth disabled
The `authCookie` example on `docs/06-concepts/04-authentication/10-web-authentication.md:21-26` has only a commented-out child:

```yaml
authCookie:
  # secure: false # Uncomment only for http://localhost development.
```

YAML parses that as `null`, and the config loader treats an empty `authCookie` map as not configured (`config.dart:1566-1582`, `_buildConfigMap`).

Verified on rc.2 by running the YAML through `ServerpodConfig.loadFromMap`: `config.authCookie` is `null`. The server keeps returning tokens in the body, and a web client with `cookieAuth = true` throws `StateError: cookieAuth is enabled but the server returned the auth token in the response body…`.

  - [ ] Add a real child to the example, as the template does (`secure: false` dev, `sameSite: lax` prod).
  - [ ] Reword "All `authCookie` fields are optional" (line 30): at least one must be present to enable the feature.
  - [ ] Add the `StateError` to the troubleshooting list (line 65).

Optionally: Make a present-but-empty `authCookie` key enable the defaults. That would be a separate framework issue.

### LINKED ISSUE #804: docs: Configuration reference omits allowedOrigins and the authCookie keys
`docs/06-concepts/lookups/configuration-reference.md:7` says it lists "every configuration option Serverpod's core library reads", but the 4.0 web-auth keys are absent. They are documented only on `06-concepts/04-authentication/10-web-authentication.md:19-41`. Source: `packages/serverpod_shared/lib/src/environment_variables.dart:289-295, 369-376`; defaults in `config.dart` (`WebAuthCookieConfig`).

| Env var | YAML key | Default |
| --- | --- | --- |
| `SERVERPOD_ALLOWED_ORIGINS` | `allowedOrigins` | none (comma-separated in env; required when `authCookie` is set) |
| `SERVERPOD_AUTH_COOKIE_NAME` | `authCookie.name` | `serverpod_auth` |
| `SERVERPOD_AUTH_COOKIE_REFRESH_NAME` | `authCookie.refreshName` | `<name>_refresh` |
| `SERVERPOD_AUTH_COOKIE_DOMAIN` | `authCookie.domain` | host-only |
| `SERVERPOD_AUTH_COOKIE_PATH` | `authCookie.path` | `/` |
| `SERVERPOD_AUTH_COOKIE_SECURE` | `authCookie.secure` | `true` |
| `SERVERPOD_AUTH_COOKIE_SAME_SITE` | `authCookie.sameSite` | `lax` |

- [ ] Add the seven rows above to the reference table, linking to the web-authentication page.
- [ ] Note that the `authCookie` section must contain at least one key to enable cookie auth (see the cookie-auth example issue).
- [ ] While there: `redis.enabled` default is listed as `false` (line 58); when a `redis` section exists it defaults to `true` (`config.dart:964`).

FXschwartz:
Three small 4.0 corrections

- **Migration tag** (#792): the upgrade guide's `--tag "upgrade-4.0"` is rejected by the CLI because tags allow only lowercase letters, numbers, and dashes. Example changed to `upgrade-4-0`; the rule is now stated on the migrations page and the `create migration` intro.
- **Cookie-auth example** (#793): the `authCookie:` block had only a commented-out child, which YAML reads as `null`, so cookie auth stayed disabled. Verified by running the YAML through `ServerpodConfig.loadFromMap` on rc.2. The example now sets `sameSite: lax`, the page says any single field enables the section, and the resulting client `StateError` is listed under verify.
- **Configuration reference** (#804): adds `allowedOrigins` and the six `authCookie.*` keys with env vars and defaults, and corrects the `redis.enabled` default (`true` when a `redis` section is present).

Markdown lint passes on all changed files.

Fixes #792, fixes #793, fixes #804.


docs/06-concepts/lookups/configuration-reference.md
```diff
@@ -55,10 +55,17 @@ Ports, hosts, and connection settings for the API, Insights, and web servers, th
...
 | SERVERPOD_REDIS_USER                             | redis.user                          | -         | The user name for Redis authentication                                                                                                                                                                       |
-| SERVERPOD_REDIS_ENABLED                          | redis.enabled                       | false     | Indicates if Redis is enabled                                                                                                                                                                                |
+| SERVERPOD_REDIS_ENABLED                          | redis.enabled                       | true      | Indicates if Redis is enabled. Defaults to `true` when a `redis` section is present; Redis is off when the section is absent or `enabled` is `false`.                                                        |
```
developerjamiu, line 58: Suggestion: this leaves `docs/06-concepts/07-operations/06-scalability.md:108` on the old default, so after merge the two pages disagree on the same fact. Worth folding into this PR:

```
| `redis.enabled` | `true` when a `redis` section is present | Required for shared cache and global events. |
```

docs/06-concepts/04-authentication/10-web-authentication.md
```diff
@@ -20,14 +20,16 @@ Enable cookie auth by adding an `authCookie` section to your server configuratio
...
 allowedOrigins:
   - https://app.example.com
 ```
 
+In `config/development.yaml`, set `secure: false` instead so the cookie is sent over `http://localhost`. The section must contain at least one field: an `authCookie:` key with no children is read as unset, and cookie auth stays off.
+
 `allowedOrigins` is required when `authCookie` is set: it backs the CSRF origin checks and credentialed CORS, which cannot use a wildcard origin. List every browser origin that calls your server. With cookie auth enabled, browsers on origins that are not in the list lose cross-origin access, including to public endpoints.
 
-All `authCookie` fields are optional:
+Every field has a default, so any single field enables the section:
```
developerjamiu: Suggestion: line 28 covers this rule already, so this could go back to just introducing the table.

```suggestion
Each field and its default:
```

docs/06-concepts/04-authentication/10-web-authentication.md
```diff
@@ -20,14 +20,16 @@ Enable cookie auth by adding an `authCookie` section to your server configuratio
 
 ```yaml
 authCookie:
-  # secure: false # Uncomment only for http://localhost development.
+  sameSite: lax
 allowedOrigins:
   - https://app.example.com
 ```
 
+In `config/development.yaml`, set `secure: false` instead so the cookie is sent over `http://localhost`. The section must contain at least one field: an `authCookie:` key with no children is read as unset, and cookie auth stays off.
```
developerjamiu: Suggestion: "instead" reads as a swap for the example's `sameSite: lax` rather than an addition for local development.

```suggestion
For local development, use `secure: false` in `config/development.yaml` so the cookie is sent over `http://localhost`. The section must contain at least one field: an `authCookie:` key with no children is read as unset, and cookie auth stays off.
```

docs/06-concepts/lookups/configuration-reference.md
FXschwartz: Good catch! Applied your row verbatim in the scalability page so both pages now state the same default.

docs/06-concepts/04-authentication/10-web-authentication.md
FXschwartz: Agreed, the rule was stated twice within a few lines. Reverted the intro to "Each field and its default:"

docs/06-concepts/04-authentication/10-web-authentication.md
FXschwartz: Agreed, "instead" implied swapping out sameSite: lax. Took your wording so it reads as an addition for local development.

developerjamiu (approved):

PR #807, opened 2026-09-09
docs: Fix includeList nested include parameter name

### LINKED ISSUE #798: docs: Relation query examples use includes: instead of include:
`docs/06-concepts/03-data-and-the-database/02-database/07-relation-queries.md:78, 86, 104` calls `Employee.includeList(includes: …)` and says "the `includes` field must be used". The generated parameter is `include`, same as on `findById`:

```dart
static OrderIncludeList includeList({
  ...
  OrderInclude? include,
})
```

Confirmed with `dart analyze` against the rc.2 test server: `includes:` fails with `The named parameter 'includes' isn't defined`; `include:` compiles. No `includes:` argument exists anywhere in `packages/` or the test server. The only `includes` in the generator is an internal getter on the IncludeList class (`model_library_generator.dart:3141`).
  - [ ] Rename `includes:` to `include:` in both examples.
  - [ ] Reword line 78: nested includes are passed through the `include` parameter, the same name used on `findById`.

Zfinix:
The nested include on `includeList` is documented as `includes`. The generated
parameter is `include`, the same name used on `findById`, so both examples on
the page fail to compile as written.

Verified against 4.0.0-rc.2, which emits `where`, `limit`, `offset`, `orderBy`,
`orderByList`, and `include`. The `versioned_docs` copies carry the same error
back to 1.2.0 and are left for a separate pass, since they are frozen release
snapshots. Closes #798.

### DIFF
--- a/docs/06-concepts/03-data-and-the-database/02-database/07-relation-queries.md
+++ b/docs/06-concepts/03-data-and-the-database/02-database/07-relation-queries.md
@@ -75,7 +75,7 @@ The example above returns a company with all related employees included.
 
 ### Nested includes
 
-The `includeList` method works slightly differently from a normal `include` and to include nested objects the `includes` field must be used. When including something on a list it means that every entry in the list will each have access to the nested object.
+The `includeList` method works slightly differently from a normal `include`. To include nested objects, use its `include` parameter. When including something on a list it means that every entry in the list will each have access to the nested object.
 
 ```dart
 var company = await Company.db.findById(
@@ -83,7 +83,7 @@ var company = await Company.db.findById(
   companyId,
   include: Company.include(
     employees: Employee.includeList(
-      includes: Employee.include(
+      include: Employee.include(
         address: Address.include(),
       ),
     ),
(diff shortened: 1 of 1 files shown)


docs/06-concepts/03-data-and-the-database/02-database/07-relation-queries.md
```diff
@@ -75,15 +75,15 @@ The example above returns a company with all related employees included.
 
 ### Nested includes
 
-The `includeList` method works slightly differently from a normal `include` and to include nested objects the `includes` field must be used. When including something on a list it means that every entry in the list will each have access to the nested object.
+The `includeList` method works slightly differently from a normal `include`. To include nested objects, use its `include` parameter. When including something on a list it means that every entry in the list will each have access to the nested object.
```
developerjamiu: Suggestion: with `includes` gone, "works slightly differently" has nothing left to hang on, and the actual difference only arrives in the next sentence. #798's second item asked for the `findById` framing, which is in the PR body but not on the page.

```suggestion
The `includeList` method takes its nested includes through an `include` parameter, the same name used on `findById`. When including something on a list it means that every entry in the list will each have access to the nested object.
```

Optional. Fine to merge as-is if you would rather keep the diff at three lines.

developerjamiu (approved):

PR #824, opened 2026-09-14
docs: Add a note on the migration guide due to the rate limiter refactor

marcelomendoncasoares:
Documentation for the change made on https://github.com/serverpod/serverpod/pull/5702.


docs/11-upgrading/01-upgrade-to-four.md
```diff
@@ -177,11 +177,17 @@ New projects use `ghcr.io/serverpod/postgres:16`, which bundles pgvector and Pos
...
+$ serverpod create-migration --tag "upgrade-4-0" --force
 ```
 
 This writes a new migration to `<project>_server/migrations/`. It will be applied to your database in the next step.
 
+:::note
+
+The `--force` flag is only needed when using the authentication module due to a small change on the table that stores rate-limit attempts. The migration is safe and won't affect existing accounts, sessions or other auth data. The only side-effect is that ongoing requests will have their attempt counters reset to zero.
+
+:::
```
developerjamiu: Every upgrader follows this step, but the note scopes `--force` to auth-module users, so as written everyone learns to pass a flag that suppresses legitimate safety warnings. Keeping the main command plain and moving the flag into the admonition keeps the teaching scoped:

````suggestion
$ serverpod create-migration --tag "upgrade-4-0"
```

This writes a new migration to `<project>_server/migrations/`. It will be applied to your database in the next step.

:::caution

With the authentication module installed, `create-migration` warns about a change to the table that stores rate-limit attempts. Add `--force` to proceed. The change is safe: accounts, sessions, and other auth data are untouched. The only side effect is that existing rate-limit counters reset to zero.

:::
````

developerjamiu (approved):
Just the optional suggestion and we are good.

PR #827, opened 2026-09-14
docs: Reference stable Serverpod 4.0

developerjamiu:
Serverpod 4.0.0 is out, so the docs now point at the stable release instead of the betas, following #822 and #823.

- Package pins in the upgrade guides, auth setup, modules, legacy auth pages and shared packages move to `4.0.0`. The private dependencies example drops its 2.x `serverpod_serialization` pin, which can't resolve in a 4.0 workspace.
- The upgrade guide installs the CLI without a version. It and the quickstart now state the Flutter 3.44.4 (Dart 3.12.2) minimum.
- The public beta wording is gone from the quickstart.


docs/06-concepts/01-server-fundamentals/04-modules.md
```diff
@@ -18,11 +18,11 @@ To add the `serverpod_auth_idp` module, add `serverpod_auth_idp_server` to your
 
 ```yaml title="myproject_server/pubspec.yaml"
 dependencies:
-  serverpod_auth_idp_server: 4.0.0-beta.0
+  serverpod_auth_idp_server: 4.0.0
 ```
 
 :::info
-Match the version to the Serverpod version you are using; all official packages share the same version number, so matching versions work together. Prerelease versions must be pinned exactly, as above; from a stable release on, a caret constraint such as `^4.0.0` works.
+Use the same version as your Serverpod packages, such as `4.0.0`. All official packages share one version number, so matching versions work together. Pin the exact version instead of a caret range. An exact pin keeps the packages in sync with the CLI.
```
FXschwartz: nit: "Pin the exact version instead of a caret range. An exact pin keeps the packages in sync with the CLI." reads as two sentences saying one thing. Could be: "Pin the exact version rather than a caret range, so the packages stay in sync with the CLI."

FXschwartz (approved):
One super minor nit but approving either way!

PR #829, opened 2026-09-14
docs: Update the 4.0 upgrade guide

developerjamiu:
The upgrade guide now matches what ships in 4.0, with the required steps first and the optional ones last.

- Breaking changes get an overview table, and cover what was missing: file storage renames, the Google web redirect setup, future calls running at least once, sign-in button styles, removed APIs, and new auth behaviors.
- The dependency steps add `serverpod_test`, the CLI deactivate step, and what to do when `serverpod generate` fails on old APIs.
- The agent setup no longer recommends `serverpod create .`, which overwrites project files. The dev workflow, production and CI sections are corrected.


FXschwartz (approved):
LGTM!

PR #830, opened 2026-09-14
docs: Update the legacy auth migration for Serverpod 4.0

developerjamiu:
Updates the legacy auth migration guide and the legacy auth pages to match shipped Serverpod 4.0 behavior.

- Migration guide: the email endpoint now runs the base login before importing legacy passwords, the `serverpod_auth_migration_client` and `googleClientSecret` requirements are added, and the legacy removal steps are in order.
- Migration guide: warns that Google, Apple, and Firebase accounts aren't linked automatically, and explains what old clients can still do while they catch up.
- Legacy Sign in with Apple: documents the required `appleClientIds` setting on the Apple page, in the Setup table, and in the Upgrade to 4.0 guide.
- Marks the bridge and migration packages as experimental, and rewords the anonymous and passkey experimental warnings to say what's still missing.


FXschwartz (approved):
LGTM!

PR #831, opened 2026-09-14
docs: Correct framework reference pages for Serverpod 4.0

developerjamiu:
Corrects the server fundamentals and CLI reference pages to match what shipped in Serverpod 4.0.

- Configuration: a Dart `ServerpodConfig` replaces the YAML files and environment variables instead of layering on top, and `configOverride` adjusts the result.
- Running your server: what each kind of saved file triggers, migrations applied on every boot, and what the `maintenance` role actually does.
- Your Serverpod project: MCP config and skills folders for each editor, and which of those files the workspace `.gitignore` excludes.
- `serverpod create`: the setup screen versus the `--no-interactive` flags, and how to create a project without a database.


FXschwartz (approved):
LGTM!

PR #832, opened 2026-09-14
docs: Update Cloud docs for Serverpod 4.0

developerjamiu:
Brings the Cloud docs in line with Serverpod 4.0 and the current scloud CLI.

- Rewrites the `scloud launch` flow: pick or create the project in the Console, the deploy is tracked inline, and the web app is served at the root.
- Updates deployment stages from four to three (Upload, Cloud build, Rollout) across the concept, CI and recovery pages.
- Corrects migration behavior: failures show in the server logs, and applied migrations are undone with a forward migration.
- Adds session log cleanup setup, and corrects the database restore note and post-deploy hook failure behavior.


Isakdl (changes requested):
I suggest we change the command to `serverpod cloud` instead of `scloud`

Isakdl (approved):
LGTM

PR #833, opened 2026-09-14
docs: Correct testing, database and file storage pages for Serverpod 4.0

developerjamiu:
Corrects the testing, database and file storage pages to match Serverpod 4.0 behavior.

- Testing: each `withServerpod` group runs against its own database, so the `--concurrency=1` advice is removed. Also adds `databaseInterceptor`, the 120-second start timeout and the `features: database: false` note.
- Database: explains what happens when a migration fails in each run mode, documents the `upsertRow` null return, and notes that pgvector and PostGIS ship with new projects. Also adds the embedded PostgreSQL cache paths and the files SQLite needs on the web.
- File storage: documents the `ServerpodCloudProvider` default in new projects, and explains that a native Google Cloud Storage public storage needs a publicly readable bucket.
- Streaming: a refused stream now fails with the same `ServerpodClientHttpException` subclasses as a regular call.


docs/06-concepts/03-data-and-the-database/02-database/16-client-side-database.md
```diff
@@ -55,8 +55,15 @@ Future<String> resolveDatabasePath(String fileName) async {
 
 Creating the session opens the database file, so keep one session for the lifetime of the app, for example in your state management, instead of creating a new one per operation. Call `close()` on the session to close the underlying database when you are done with it. On the web there is no file system, which is why the example passes a bare name instead of a path.
 
+:::note
+On the web, SQLite runs as WebAssembly in a web worker. Your Flutter app's `web/` directory needs two files, which the [`sqlite_async` web setup](https://pub.dev/packages/sqlite_async#web) links to:
+
+- `sqlite3.wasm`, from the release that matches the `sqlite3` version in your `pubspec.lock`. Without it, `createSession` fails.
+- `db_worker.js`, from the release that matches the `sqlite_async` version in your `pubspec.lock`. Without it, the database runs on the page's main thread and is not safe to use from several tabs.
```
FXschwartz, line 62: I think `markdownlint` is going to trip on this one (MD032, blanks around lists). The list runs straight into the closing `:::` on the next line. Could we add a blank line after the `db_worker.js` bullet? I ran the repo config locally and this was the only hit.

docs/06-concepts/03-data-and-the-database/02-database/13-migrations.md
```diff
@@ -134,7 +139,7 @@ If migrations are applied at the same time as repair migration, the repair migra
 
 ### On Serverpod Cloud
 
-Serverpod Cloud applies migrations for you. Every deploy starts the server with `--apply-migrations`, so any pending migrations run before the server serves requests. If a migration fails, the deploy fails; fix the migration and redeploy. See [Cloud database](/cloud/concepts/database#migrations-run-on-every-deploy) for the full flow.
+Serverpod Cloud applies pending migrations for you on every deploy. If a migration fails, the server logs `Failed to apply database migrations.` Fix the migration and redeploy. See [Cloud database](/cloud/concepts/database#migrations-run-on-every-deploy) for the full flow.
```
FXschwartz, line 142: This paragraph now says a failed migration on Cloud just gets logged, but the Cloud page it links to still says the server fails to start and the deploy is reported as failed. The server side matches what you wrote (outside development it logs and keeps going), so I'm not sure which page is right about what the deploy pipeline does.

Might be worth checking with the Cloud folks and updating both pages together so they don't disagree with each other.

FXschwartz (changes requested):
Just a couple of comments. As a side note not something for this PR, but two of the things you documented look more like framework gaps than intended behavior: `--no-database` not writing `features: database: false`, and the maintenance role exiting 0 after a failed migration.

I'll bring it up for discussion during standup tomorrow @marcelomendoncasoares

docs/06-concepts/03-data-and-the-database/02-database/16-client-side-database.md
developerjamiu: Applied.

docs/06-concepts/03-data-and-the-database/02-database/13-migrations.md
developerjamiu: #832 updates the Cloud side to match: the database page and the failed-deploy guide drop "fails to start" and "reported as failed", and point to `Failed to apply database migrations.` in `scloud log`. What the deploy pipeline reports isn't in the Cloud or framework source we can check, so neither page claims it. If the Cloud team confirms, we can add it to both in one follow-up.

FXschwartz (approved):

PR #835, opened 2026-09-14
docs: Correct the hosting elsewhere page for Serverpod 4.0

developerjamiu:
Updates the hosting elsewhere page to match how Serverpod 4.0 creates, builds and runs projects.

- Replaces the removed Serverpod mini exception with projects created without a database, for example with `serverpod create --no-interactive --no-database`.
- Adds the `docker build` command, which must run from the project root so the build can copy `pubspec.lock`.
- Corrects the `runmode` default in the Docker table to `production`, and marks it as the `Dockerfile` default.


FXschwartz (approved):
LGTM!

PR #836, opened 2026-09-15
docs: Clarify the SQLite logging and Redis defaults text

developerjamiu:
Rewrites two paragraphs on the server fundamentals pages so they read clearly and no longer contradict themselves.

- Redis: replaces the paragraph that said Redis is included "only if you opted in" but also "by default" with a short list of how each create path decides (setup screen, `--no-interactive`, `serverpod quickstart`).
- SQLite: splits the session logging paragraph into short sentences. No facts change.


FXschwartz (approved):
LGTM!

PR #847, opened 2026-09-15
docs: Document MCP editor setup and migration safety

Zfinix:
## Summary

The `serverpod mcp-server` page documented the `--server-dir` flag and nothing else, so there was nowhere to send a reader who wants to connect an agent to their project. It also said nothing about `force`, which the migration tools accept and an agent can pass on its own.

This PR:

- Adds a "Set up your editor" section listing the six editors `serverpod create` configures and the config file each one gets.
- Adds a "Set up any other MCP client" section with the two commands Serverpod registers, so clients outside that list can be wired up by hand.
- Adds a "Migrations and your data" section warning that an agent decides on its own when to pass `force`, and to point the bridge at a development database.

The tool reference itself is generated and arrives separately through serverpod/serverpod#5711. Only the hand-maintained `_mcp-server.md` intro changes here, which the docs-sync generator preserves across runs.


docs/06-concepts/cli/commands/mcp-server/_mcp-server.md
```diff
@@ -3,3 +3,56 @@
...
+
+VS Code names the block `servers` rather than `mcpServers`, Antigravity names the Dart entry `dart-mcp-server`, and Codex uses TOML. The two commands are the same in every case.
+
+:::note
+Module projects get only the Dart MCP server. A module has no runnable server, so there is no `serverpod start` session for the bridge to connect to.
+:::
+
+## Migrations and your data
+
+The migration tools change your database. Both `create_migration` and `create_repair_migration` accept a `force` parameter, which proceeds even when Serverpod warns that data may be destroyed.
```
developerjamiu: The bridge exposes a third migration tool, and it is the one that actually changes the database: `apply_migrations`. The create tools only write migration files, and their `force` is the gate that authorizes a destructive migration (its description says "Required for destructive migrations"). Naming all three keeps the safety story complete, since an agent can generate and apply end to end:

```suggestion
The bridge exposes three migration tools. `create_migration` and `create_repair_migration` write migration files without touching the database. Both accept a `force` parameter, which is required when the migration would destroy data. `apply_migrations` then applies the pending migrations to the running server's database.
```

Zfinix: Updated to name all three, and the warning now points at `apply_migrations`.

developerjamiu (approved):

PR #848, opened 2026-09-16
docs: Correct Cloud docs details against Cloud CLI 1.0.0

developerjamiu:
Accuracy corrections in the Cloud docs, verified against Cloud CLI 1.0.0.

- Scheduled snapshots without `--retention` are kept for 35 days, not 24 hours, and 35 days is also the maximum. Adds a note that a schedule set before that change still expires its snapshots after 24 hours.
- Sample tables now show the time zone labels the CLI prints on timestamp columns.
- The GitHub Actions troubleshooting entry quotes the message the CLI actually prints, and adds a revoked token as a cause.
- The `.gitignore` sample matches what the CLI writes, and the page now says the CLI adds it for workspace projects.


cloud_docs/concepts/database-backups.md
```diff
@@ -27,7 +27,13 @@ A schedule takes snapshots for you at a fixed frequency. Each database has one s
...
+The other options control when the schedule runs and how long its snapshots live:
+
+- **`--day`** picks the day of the week (1-7) for a weekly schedule, or the day of the month (1-31) for a monthly one. It defaults to 1 and does not apply to a daily schedule.
+- **`--hour`** sets the hour of the day, from 0 to 23 in UTC. It defaults to 0.
+- **`--retention`** sets how long each scheduled snapshot is kept before it is deleted automatically, for example `30d`. Without it, snapshots are kept for 35 days, which is also the longest a schedule can keep them.
+
+The 35-day default applies to new schedules only. A schedule created without `--retention` before it still expires its snapshots after 24 hours, so run the command again with a `--retention` value to change that.
```
SandPod: Question: I don't understand this section. Above, we just said that they are always kept for 35 days if `--retention` is omitted. Connect the two instructions or make the difference clearer, please.

SandPod (changes requested):
I have one request for clarification.

cloud_docs/concepts/database-backups.md
developerjamiu: Adjusted. The default changed from 24 hours to 35 days and existing schedules were not updated, which the page never said. Both cases now sit in the `--retention` bullet.

SandPod: Looks better, but perhaps break out the old default thing into an info bubble and give it a date so that the user understands if he is affected or not.

developerjamiu: Done. The old default is now an info bubble with the date it changed, 15 September 2026, so a reader can tell whether their schedule predates it.

cloud_docs/concepts/database-backups.md
```diff
@@ -27,7 +27,15 @@ A schedule takes snapshots for you at a fixed frequency. Each database has one s
...
+The other options:
+
+- **`--day`** sets the day of the week (1-7) for a weekly schedule, or the day of the month (1-31) for a monthly one. It defaults to 1 and does not apply to a daily schedule.
+- **`--hour`** sets the hour of the day, from 0 to 23 in UTC. It defaults to 0.
+- **`--retention`** sets how long each scheduled snapshot is kept before it is deleted automatically, for example `30d`. Without it, snapshots are kept for 35 days, which is also the longest allowed.
+
+:::info
+Serverpod Cloud changed this default on 15 September 2026. A schedule set without `--retention` before then still keeps its snapshots for 24 hours. Run `serverpod cloud db schedule show` to see which retention yours uses, and set the schedule again with a `--retention` value to change it.
```
SandPod:
```suggestion
The default retention was changed on 15 September 2026. A schedule set without `--retention` before then still keeps its snapshots for 24 hours. Run `serverpod cloud db schedule show` to see which retention yours uses, and set the schedule again with a `--retention` value to change it.
```

SandPod (approved):
Last suggestion, otherwise it looks good! 👍

PR #849, opened 2026-09-16
docs: Add intros for the build and storage command pages

developerjamiu:
Fills in the two empty hand-written intros for the CLI reference pages added with the `build` and `storage` commands.

- `scloud build`: frames the build log as the build-time error surface, separate from the runtime logs `scloud log` fetches, and points at Deployments, Recover from a failed deploy, and Private dependencies.
- `scloud storage`: says what a storage is for, that access is fixed at creation, and links the framework's File uploads page for the server-side side of it.
- Both verified against the CLI source at 1.0.0, and neither repeats the generated usage block below it.


cloud_docs/reference/cli/commands/storage/_storage.md
```diff
@@ -0,0 +1,7 @@
+# scloud storage
+
+The `scloud storage` commands work on your project's file storage from the terminal, both the storages themselves and the files inside them. Your server reads and writes the `public` and `private` storages through `session.storage`. That makes `scloud storage file list public` a view of what your app uploaded, and `scloud storage file upload` a way to put a file in place before anyone uses the app.
```
Isakdl: Suggestion: `public` and `private` here are storage IDs, that happen to be the default ones we provisioned. I would word this more generic considering this is the reference documentation.

Isakdl (changes requested):
Nice catch! Just have a smaller nit comment

cloud_docs/reference/cli/commands/storage/_storage.md
developerjamiu: Applied

Isakdl (approved):
LGTM

PR #850, opened 2026-09-16
docs: Correct what Serverpod Cloud manages for storage and Redis

developerjamiu:
Managed file storage shipped in Cloud CLI 1.0.0, so the overview bullet that grouped it with pub/sub and caching as bring-your-own is out of date. A user hit this after reading the pricing page and finding the docs said the opposite.

- Splits the bullet: file storage is managed and every project starts with a private and a public storage, while pub/sub and caching stay bring-your-own Redis.
- Corrects the Redis page in both the current and 4.0.0 docs, which described the Cloud Redis guide as "managed Redis on Serverpod Cloud".
- Drops the "not yet" hedge from the Cloud Redis guide, matching the wording the Caching page already uses.


developerjamiu:
The File storage bullet links to the CLI reference page because there is no Cloud storage concept page yet. #771 adds one, so it would be good to merge that soon and repoint this link at the concept page, the way the other bullets in that list point at concepts.

SandPod (approved):
👍

Zfinix:
LGTM

PR #851, opened 2026-09-17
docs: Correct the maintenance role exit code after a failed migration

FXschwartz:
serverpod/serverpod#5709 restores the exit code for the maintenance role, so a failed migration now exits with code 1 in every run mode. This reverts the wording from #833 that described exiting with code 0 outside `development` as expected behavior.

- Migrations: the maintenance role paragraph now says a failed or mismatched migration exits with code 1.
- Running your server: same correction in the server role list.


developerjamiu:
Suggestion: apply the same two changes to the 4.0.0 versioned docs as well, since that's what readers see by default and new projects already pin 4.0.1. Both files carry the old sentence:

- `versioned_docs/version-4.0.0/06-concepts/01-server-fundamentals/02-running-your-server.md` line 117
- `versioned_docs/version-4.0.0/06-concepts/03-data-and-the-database/02-database/13-migrations.md` line 132

developerjamiu:
Nit: in `docs/06-concepts/03-data-and-the-database/02-database/13-migrations.md` line 231, the repair-migration paragraph only says the server exits after applying. The new exit code covers `--apply-repair-migration` too, and line 132 now says so for migrations. Worth matching so the two maintenance paragraphs agree:

> The repair migration can also be applied with the [maintenance role](../../server-fundamentals/running-your-server#choose-a-server-role), where the server exits after applying it, with code 1 if it fails.

developerjamiu (approved):

developerjamiu:
Did you checked the comments @FXschwartz ?

FXschwartz:
@developerjamiu Sorry about that! I saw the approval but completely missed your comments before merging. I will open a new PR to handle your comments. 

Apologies!

developerjamiu:
No worries. I actually thought you might have missed the comments. That was why I sent the message to check.

